PR #2, opened 2023-03-16
Rename f- to w-

AnnaRybkina:


pearofducks:
This looks fine, but elements like `<w-docs-template>` don't exist - so they probably need removed instead of renamed? I don't know if y'all have discussed this or not.

pearofducks (approved):

AnnaRybkina:
> This looks fine, but elements like `<w-docs-template>` don't exist - so they probably need removed instead of renamed? I don't know if y'all have discussed this or not.

This PR is a good place to discuss such things. Please do a code change suggestion in the comment to visualize what you mean.

pages/components/alert.html
```diff
@@ -1,7 +1,7 @@
 <html lang="en">
   <%- include('head.html'); -%>
   <body>
-    <f-docs-template>
+    <w-docs-template>
```
pearofducks, line 4: This component - and any others from [here](https://github.com/fabric-ds/common/tree/main/packages) won't be ported over to Warp.

BalbinaK: tl;dr
I think for now you can leave the whole `pages` directory as is, until we have time to discuss what warp elements documentation will look like.

The original elements repo would add pages to the [fabric documentation website](https://elements.fabric-ds.io/alert.html) for each component, and it would use a web component like `f-docs-template` that adds styles, header and sidebar to a page. The templates are imported in [scripts.html](https://github.com/warp-ds/elements/blob/main/pages/includes/scripts.html#L1-L4). Because they rely heavily on fabric and we're building a new documentation site, we don't need this custom wrapper in the Warp docs. There are more things in each component page that rely on fabric in this repo, and I'm not sure what we'd like to keep for later, and what would just be forgotten. That's why I think we could wait with touching the whole `pages` directory and see what we do about it when the time comes. Wdyt?

AnnaRybkina: I reverted the useless renaming https://github.com/warp-ds/elements/pull/2/commits/32be3068535af4dfc155bfa33c0db28142671546. Let's handle it later

BalbinaK (approved):

PR #3, opened 2023-03-16
Remove all css from Fabric and use uno/warp/drive

AnnaRybkina:
Somewhat working, right?

image


pearofducks (approved):

AnnaRybkina:
@pearofducks I seem to have troubles with compiling ts. Seems to be problems with how pnpm installs some of the packages and then ts try to compile stuffs in libs. Maybe the symlinking is off somehow fort these packages. Do you recognize the problem/behaviour? Everything works with npm.

I've tried a lot of things, e.g. adding `preserveSymlinks`, `typeRoots` etc, to the tsconfig with no luck. The only one thing that works is `"skipLibCheck": true` but maybe you know another way. 
image

pearofducks:
When you say 'everything works with npm' is that install and running the script? Or just running the script?

pnpm is more strict about peer dependencies, and has very minor changes in running scripts - but otherwise it should be basically the same.

pearofducks:
Working with this in tsconfig

```
{
  "include": ["index.js"],
  "compilerOptions": {
    "module": "esnext",
    "lib": ["ES2020", "DOM", "DOM.Iterable"],
    "allowJs": true,
    "declaration": true,
    "emitDeclarationOnly": true,
    "outDir": "dist",
    "moduleResolution": "node",
    "preserveSymlinks": true,
    "allowSyntheticDefaultImports": true,
    "skipLibCheck": true
  }
}
```

AnnaRybkina:
> Working with this in tsconfig
> 
> ```
> {
>   "include": ["index.js"],
>   "compilerOptions": {
>     "module": "esnext",
>     "lib": ["ES2020", "DOM", "DOM.Iterable"],
>     "allowJs": true,
>     "declaration": true,
>     "emitDeclarationOnly": true,
>     "outDir": "dist",
>     "moduleResolution": "node",
>     "preserveSymlinks": true,
>     "allowSyntheticDefaultImports": true,
>     "skipLibCheck": true
>   }
> }
> ```
> 
> (I also cleaned up the dependencies locally here because they're in a pretty ugly place, but I think it was the tsconfig change that did it)

Ok, so my suggestion about `"skipLibCheck": true` is the way to go then. `preserveSymlinks` didn't do any difference when I tested

pearofducks:
This is not necessarily in the scope of this PR, but nothing should be importing from `lit-element` or `lit-html` anymore, and those dependencies should be gone from the project. `lit` itself brings those in, and re-exports their exports. e.g. `unsafeHTML` - https://lit.dev/docs/api/directives/#unsafeHTML

PR #4, opened 2023-03-16
Add release step

AnnaRybkina:


.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
+name: Release
+on:
+  push:
+    branches:
+      - main
+      - alpha
```
BalbinaK, line 6: We added `beta` just in case in other repos, so maybe we could add it here, too?

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+      - next
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
```
BalbinaK, line 13: Checkout repository 🙃

.releaserc.json
```diff
@@ -0,0 +1,15 @@
+{
+  "branches": [
+    { "name": "main" },
+    { "name": "alpha", "prerelease": true },
+    { "name": "next", "prerelease": true }
```
BalbinaK, line 5: add beta here, too

package.json
```diff
@@ -23,8 +24,8 @@
     "lint:eslint": "eslint . --ext js,cjs,mjs --ignore-path .gitignore",
     "dev": "vite",
     "build:docs": "vite build && cp CNAME site",
-    "eik:login": "eik login",
-    "eik:publish": "eik publish",
+    "eik:login": "npx @eik/cli login",
+    "eik:publish": "npx @eik/cli publish",
     "eik:alias": "eik pkg-alias",
```
BalbinaK, line 29: I don't think this is used anywhere, so I guess we can remove it.

BalbinaK (commented):
Did you intend to remove pnpm-lock.yaml?

package.json
AnnaRybkina: Don’t we use them here https://github.com/warp-ds/elements/pull/4/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R32

.github/workflows/release.yml
AnnaRybkina: Sure 👍🏻

.github/workflows/release.yml
AnnaRybkina: Ah good catch! I fix 🙌🏻

PR #10, opened 2023-04-12
feat(workflows): add eik aliasing script to release.yml

BalbinaK:
Allow the latest version of the package to be imported from eik server without the need to update the version number every time a new bundle version is published.
`https://assets.finn.no/pkg/@warp-ds/vue/v1/index.js` 
instead of
`https://assets.finn.no/pkg/@warp-ds/vue/v1.0.0-alpha.9/index.js`

More info on aliases in eik can be found [here](https://eik.dev/docs/client_aliases/).

When `1.0.0` version is published to NPM (`alpha` is merged to `main`), the `eik pkg-alias` script should only be run on pushes to `main`, which will be handled in [this task](https://trello.com/c/IfLXuWlg).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,6 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: npx semantic-release
       - name: Eik login and publish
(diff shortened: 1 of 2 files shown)


AnnaRybkina (approved):
Sorry, missed this one 🤦

PR #11, opened 2023-04-17
fix: add styles to textfield

AnnaRybkina:
Can be tested on http://localhost:xxxx/pages/components/textfield.html

### DIFF
--- a/packages/textfield/index.js
+++ b/packages/textfield/index.js
@@ -1,4 +1,5 @@
 import { css, html, LitElement } from 'lit';
+import { input, label as l, helpText as h } from '@warp-ds/component-classes';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { fclasses } from '../utils';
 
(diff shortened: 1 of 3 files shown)


packages/textfield/index.js
```diff
@@ -1,4 +1,5 @@
 import { css, html, LitElement } from 'lit';
+import { input, label as l, helpText as h } from '@warp-ds/component-classes';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { fclasses } from '../utils';
```
AnnaRybkina, line 4: Shall we rename this one? To something else than `fclasses`. Maybe `classes` || `componentClasses` or something more informative?
```suggestion
import { componentClasses } from '../utils';
```

BalbinaK: What about something like `mapClasses`? I see its main role is to map classes based on an object we give it and use `lit`'s directive [classMap](https://lit.dev/docs/templates/directives/#classmap) on it which will return a string of utility classes that are supposed to be applied. But no strong opinions here.

On a side note:
There is another util function called `classes` inside `../utils` that has a similar way of creating the string of class names from an object. The main difference here is that this util doesn't override already applied pairs of key-value before it makes a string out of it. 
I'll try to explain on an example:
```
const definition = { "p-4": true, "m-12": false, "border-2 p-4": false };
const classes1 = classes(definition);
const classes2 = mapClasses(definition);
console.log(classes1); // "p-4"
console.log(classes2); // "" <-- initial "p-4" key value (`true`) was overridden with `false` when that key was met again
```

There are 2 components (toast and attention) that use `classes` util now. I'm afraid they could potentially break if we decided to use `mapClasses` on them, too. The potential to have just one util is big but I haven't tested which of the two utils are more predictable. I wouldn't touch that because we wouldn't know if the problems stem from the migration to Warp or that small change.

BalbinaK (approved):
LGTM!

packages/textfield/index.js
AnnaRybkina: Tnx for your detailed explanation 🙌 I agree with you about not touching if if it is too many unknowns. I reacted primarily at the `f` letter in the name as we're renaming `f` to `w`.

PR #14, opened 2023-04-24
fix: add Warp component-classes to the breadcrumbs component

BalbinaK:
## Description
- update the breadcrumbs component to use [component-classes](https://github.com/warp-ds/component-classes/blob/alpha/index.js#L310-L315)
- color tokens work
- handle nested arrays of children like in React and Vue
- allow for overriding the default `aria-label` value

## Additionally
- render `string`-typed children as styled `span` elements with `aria-current="page"` on last child

Screenshot 2023-04-21 at 10 58 51

### DIFF
--- a/packages/breadcrumbs/index.js
+++ b/packages/breadcrumbs/index.js
@@ -1,13 +1,23 @@
 import { html, LitElement, css } from 'lit';
 import { interleave } from '@warp-ds/core/breadcrumbs';
 import { breadcrumbs as ccBreadcrumbs } from '@warp-ds/component-classes'
+import { kebabCaseAttributes } from '../utils';
 
 const separator = html`<span class=${ccBreadcrumbs.separator} aria-hidden='true'>/</span>`;
 
-class WarpBreadcrumbs extends LitElement {
+class WarpBreadcrumbs extends kebabCaseAttributes(LitElement) {
   static styles = css`
     /* @unocss-placeholder */
   `;
+
+  static properties = {
+    ariaLabel: { type: String },
+  };
+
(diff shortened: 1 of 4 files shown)


BalbinaK:
[Warpify Breadcrumbs component](https://trello.com/c/WJGrusEL/179-warpify-breadcrumbs-component)

packages/breadcrumbs/index.js
```diff
@@ -1,25 +1,48 @@
 import { html, LitElement, css } from 'lit';
 import { interleave } from '@warp-ds/core/breadcrumbs';
+import { breadcrumbs as ccBreadcrumbs } from '@warp-ds/component-classes'
+import { kebabCaseAttributes } from '../utils';
 
-const separator = html`<span class="select-none" aria-hidden="true">/</span>`;
+const separator = html`<span class=${ccBreadcrumbs.separator} aria-hidden='true'>/</span>`;
```
magnuh, line 6: Shouldn't this be with double quotes?
```suggestion
const separator = html`<span class="${ccBreadcrumbs.separator}" aria-hidden="true">/</span>`;
```

BalbinaK: As far as I checked, it's not necessary to have quotes when passing a JS value to the attributes in Web Components. 

Screenshot 2023-04-25 at 08 27 45

As for single vs. double quotes, there are no eslint rules set in this repo yet, so the inconsistency is in more files. If we want to fix this, maybe we could do it as a separate task? 😊

CarolineGard (approved):
Nice job, works well!

PR #17, opened 2023-05-04
feat(alert): warpify alert component

CarolineGard:
Warpify alert component

### DIFF
--- a/packages/alert/index.js
+++ b/packages/alert/index.js
@@ -1,35 +1,13 @@
 import { css, html, LitElement } from 'lit';
+import { alert as ccAlert } from '@warp-ds/component-classes';
 import { infoSvg, negativeSvg, positiveSvg, warningSvg } from './svgs';
+import { classNames } from '@chbphone55/classnames';
 
-const variants = {
-  negative: {
-    color: 'red',
-    icon: negativeSvg(),
-    bgColor: 'bg-red-50',
-    borderColor: 'border-red-300',
-    textColor: 'text-red-600',
-  },
-  positive: {
-    color: 'green',
-    icon: positiveSvg(),
-    bgColor: 'bg-green-50',
-    borderColor: 'border-green-300',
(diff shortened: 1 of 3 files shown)


packages/alert/index.js
```diff
@@ -47,44 +25,56 @@ class WarpAlert extends LitElement {
...
-        <div
-          role=${this.role}
-          class="${`flex p-16 border rounded-4 border-l-4 ${bgColor} ${borderColor}`}"
-          style="border-left-color:var(--w-${color}-600)"
-        >
-          <div class="mr-8 ${textColor}">${icon}</div>
-          <div class="text-14">
+        <div role=${this.role} class=${this._wrapperClasses}>
+          <div class=${this._iconClasses}>${this._icon}</div>
+          <div class="last:mb-0 text-14">
```
BalbinaK, line 77: Ah, these classes should also be exported from component-classes and updated. Could you add a TODO here so we have some way of remembering it? 🙏

packages/alert/index.js
```diff
@@ -1,35 +1,15 @@
+// TODO: replace text-14 with a token
```
BalbinaK, line 1: The thing we need to do about the hard-coded classes is not to replace them with a token but to import them from `@warp-ds/component-classes`. This way they are included in a list of component classes that is later feeded to unocss configuration's `safelist`. That makes the engine generate css for those classes at build step, which we then can pass to components which are nested within shadow DOM. Otherwise those two classes would be missing.

On another note, `text-14` is not going to be replaced with a token, which I understand may have been unclear 😄  It's a utility class that unocss will generate respective css for, and because we have recently made both `text-{number}` and `text-{t-shirt-sizes}` generate respective font-size and line-height css values, this will work. We can have a chat about it so I can shed some more light on this 😊 

Finally, it would be great if we could add TODO comments right above related lines of code, if possible 🙏

PR #21, opened 2023-05-25
feat(attention): warpify attention

CarolineGard:
Warpify attention


packages/attention/index.js
```diff
@@ -135,23 +135,25 @@ class WarpAttention extends kebabCaseAttributes(LitElement) {
...
-      [`-top-8`]: this._arrowDirection === 'top',
-      [`-right-8`]: this._arrowDirection === 'right',
-      [`-bottom-8`]: this._arrowDirection === 'bottom',
-      [`-left-8`]: this._arrowDirection === 'left',
-      [c.arrowTooltip]: this.tooltip,
-      [c.arrowCallout]: this.callout,
-      [c.arrowPopover]: this.popover,
+      [ccAttention.arrowBase]: true,
+      [ccAttention[`arrowDirection${this._arrowDirection.charAt(0).toUpperCase() + this._arrowDirection.slice(1)}`]]: true,
+      [arrowType()]: this.tooltip || this.callout || this.popover,
```
magnuh, line 156: Would have preferred the way it was before here...? 🙈😅 A lot clearer to me. Was there anything problematic with that? 🙂

pages/components/attention.html
```diff
@@ -87,7 +87,7 @@ <h3 class="mt-24 mb-16">Popover</h3>
           <w-attention placement="bottom" popover>
             <button
               id="target"
-              class="group block relative break-words last-child:mb-0 p-16 rounded-8 bg-aqua-50"
+              class="group block relative break-words last:mb-0 p-16 rounded-8 bg-aqua-50"
```
magnuh, line 90: Do you know if there is a reason for the `group` class here? 🤔

CarolineGard: haha yeah, I dunno. that's old stuff 🤔
we should clean up these docs later as we don't use them as docs docs😄

magnuh (approved):

BalbinaK (approved):

PR #22, opened 2023-05-31
fix(alert): use imported instead of inline classes

BalbinaK:
Depends on [respective @warp-ds/component-classes changes](https://github.com/warp-ds/component-classes/pull/71) and [last-child class support](https://github.com/warp-ds/drive/pull/110)


BalbinaK:
[Move inline classes from components to component-classes](https://trello.com/c/2TpB4fSE/261-move-inline-classes-from-components-to-component-classes)

CarolineGard (approved):
🙌

PR #28, opened 2023-06-19
fix(expandable): fix chevron issue

CarolineGard:
Fix issue with wrong alignment of the chevron

https://trello.com/c/WrlChPyH/273-expandable-fix-chevron-line-height-issue

### DIFF
--- a/packages/expandable/index.js
+++ b/packages/expandable/index.js
@@ -89,9 +89,9 @@ class WarpExpandable extends kebabCaseAttributes(LitElement) {
               })}
               @click=${() => (this.expanded = !this.expanded)}
             >
-              <div class="flex justify-between align-center">
+              <div class=${ccExpandable.title}>
                 ${this.title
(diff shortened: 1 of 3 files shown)


packages/expandable/index.js
```diff
@@ -89,9 +89,9 @@ class WarpExpandable extends kebabCaseAttributes(LitElement) {
               })}
               @click=${() => (this.expanded = !this.expanded)}
             >
-              <div class="flex justify-between align-center">
+              <div class=${ccExpandable.title}>
```
magnuh, line 92: Shouldn't this be quoted?
```suggestion
              <div class="${ccExpandable.title}">
```
...or is that automagically fixed someway?

CarolineGard: don't know, It did work! but I added the quotes now anyway😅

magnuh (approved):

PR #31, opened 2023-06-30
fix: use css packages instead of component-classes

AnnaRybkina:
Fixes [Warp-185](https://nmp-jira.atlassian.net/browse/WARP-185?atlOrigin=eyJpIjoiNTI4YmYyMzExYzJmNGQ0ZDkyYzg0YmVlODUzYTMzZWMiLCJwIjoiaiJ9)

In this PR we are supporting the new semantic classes that was a breaking change in the @warp-ds/css@1.0.0-alpha.33.

Changes include:

- bump @warp-ds/uno to -alpha.60 & @warp-ds/css to -alpha.33
- import component classes from @warp-ds/css/component-classes
- use new semantic color tokens from @warp-ds/css/1.0.0-alpha.33
- use new resets in uno config

Now that we have own new css repo which contains component-classes we need to update dependencies and imports

## Notes
All components in this repo are getting shipped together with all css needed. We are safelisting the component classes here. I wonder if we want to import the components.css here instead? 🤔 To get cached stylesheet. 

I'm trying to think if there are some disadvantages for us. As we will remove the safelist part we will not be able to link our local css package to test the newest changes locally as components.css won't be updated. That the only one minus I can come up with now

### DIFF
--- a/elements.css
+++ b/elements.css
@@ -1,1 +1,1 @@
-@import "https://assets.finn.no/pkg/@warp-ds/tokens/v1/finn-no.css";
+@import "https://assets.finn.no/pkg/@warp-ds/css/1.0.0-alpha.33/tokens/finn-no.css";
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@fabric-ds/icons": "0.6.7",
     "@open-wc/testing": "3.2.0",
     "@warp-ds/core": "1.0.0",
-    "@warp-ds/uno": "^1.0.0-alpha.58",
+    "@warp-ds/uno": "^1.0.0-alpha.60",
     "@warp-ds/css": "1.0.0-alpha.33",
     "glob": "8.1.0",
     "html-format": "1.1.2",
--- a/package.json
+++ b/package.json
@@ -43,8 +43,6 @@
     "@eik/cli": "^2.0.22",
     "@semantic-release/changelog": "6.0.3",
     "@semantic-release/git": "10.0.1",
-    "@warp-ds/uno": "^1.0.0-alpha.45",
-    "@warp-ds/component-classes": "1.0.0-alpha.110",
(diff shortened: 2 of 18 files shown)


BalbinaK:
> All components in this repo are getting shipped together with all css needed. We are safelisting the component classes here. I wonder if we want to import the components.css here instead? 🤔 To get cached stylesheet.
> 
> I'm trying to think if there are some disadvantages for us. As we will remove the safelist part we will not be able to link our local css package to test the newest changes locally as components.css won't be updated. That the only one minus I can come up with now

Correct me if I misunderstood something. I think that if we import components.css from Eik here, we'd need to use the exact version in the url (not a `v1`) to make sure it's in line with the dependency version we use to apply utility classes. Then whenever a new version is published to eik and npm (which happens upon each merge in @warp-ds/css), we'd need to manually update both the url and the dependency to make sure they are aligned.

AnnaRybkina:
> > All components in this repo are getting shipped together with all css needed. We are safelisting the component classes here. I wonder if we want to import the components.css here instead? 🤔 To get cached stylesheet.
> > I'm trying to think if there are some disadvantages for us. As we will remove the safelist part we will not be able to link our local css package to test the newest changes locally as components.css won't be updated. That the only one minus I can come up with now
> 
> Correct me if I misunderstood something. I think that if we import components.css from Eik here, we'd need to use the exact version in the url (not a `v1`) to make sure it's in line with the dependency version we use to apply utility classes. Then whenever a new version is published to eik and npm (which happens upon each merge in @warp-ds/css), we'd need to manually update both the url and the dependency to make sure they are aligned.

Yes, that is why we probably need to alias at some point. Otherwise it would be too difficult for us to keep up with the stuff

siljec (approved):
LGTM  🎉

PR #32, opened 2023-07-07
fix: stable FINN prod alpha-release

AnnaRybkina:
## Background
With this setup @warp-ds/elements will be compatible with [tokens](https://assets.finn.no/pkg/@warp-ds/tokens/v1/finn-no.css) currently available in FINN production. After merging this PR and so releasing a new alpha version, we will update [Warp Tech Docs](https://warp-ds.github.io/tech-docs/getting-started/developers/) and suggest the use of this particular version when relying on the above mentioned tokens.
 
## Changes
- bump `@warp-ds/uno` to `1.0.0.-alpha.49`
- bump `@warp-ds/component-classes` to `1.0.0.-alpha.116`

Also:
- Fix missing transparent background on buttons in Button of type "link" and suffix for Textfield

Before|After
---|---
image|image
image|image

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "@open-wc/testing": "3.2.0",
     "@warp-ds/core": "1.0.0",
     "@warp-ds/uno": "^1.0.0-alpha.49",
-    "@warp-ds/component-classes": "^1.0.0-alpha.115",
+    "@warp-ds/component-classes": "^1.0.0-alpha.116",
     "glob": "8.1.0",
     "html-format": "1.1.2",
     "lit": "2.7.5"
--- a/package.json
+++ b/package.json
@@ -43,8 +43,6 @@
     "@eik/cli": "^2.0.22",
     "@semantic-release/changelog": "6.0.3",
     "@semantic-release/git": "10.0.1",
-    "@warp-ds/uno": "^1.0.0-alpha.45",
-    "@warp-ds/component-classes": "1.0.0-alpha.110",
     "autoprefixer": "10.4.14",
     "cors": "2.8.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -74,6 +72,8 @@
     "@fabric-ds/icons": "0.6.7",
     "@open-wc/testing": "3.2.0",
     "@warp-ds/core": "1.0.0",
+    "@warp-ds/uno": "^1.0.0-alpha.49",
+    "@warp-ds/component-classes": "^1.0.0-alpha.116",
```
siljec, line 76: these are moved from devDependenices to dependencies, is it intentional? :)

BalbinaK: Yes, the idea was to prevent [renovate bot](https://docs.renovatebot.com/) from auto-merging patch-version updates in case we implement this tool in the same way as it was done in [Fabric](https://github.com/fabric-ds/elements/blob/fcf8315a8710ea7488033495fa7b73f2fd427d7d/.github/renovate.json#L4). We were thinking that changes from @warp-ds packages might require some testing before version updates are merged in component packages.

On a side note, I was also thinking that eventually these dependencies could be treated as peer-dependencies so that we let users know which future releases are compatible with which versions of other packages, but I haven't yet looked too much into this subject (just found [this article](https://code-trotter.com/web/dependencies-vs-devdependencies-vs-peerdependencies/)).

BalbinaK: Actually, it seems the peer-dependency warning works fine without us doing anything :D Example from installing the newest Warp React in @warp-ds/tech-docs:
Screenshot 2023-07-10 at 13 42 30

PR #33, opened 2023-07-10
fix(select): add component classes to select

BalbinaK:
This was supposed to be done together with corresponding changes from https://github.com/warp-ds/vue/pull/18 so before we announce a stable FINN prod version of this package, this had to be fixed, too.
 
Before:
Screenshot 2023-07-10 at 12 16 14

After:
Screenshot 2023-07-10 at 12 12 19

Also:
- apply unocss to the internal docs document for easier testing
- fix a bug where w-card's hidden button was visible

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -77,7 +77,10 @@ export default ({ mode }) => {
         presets: [presetWarp()],
         mode: 'shadow-dom',
         safelist: classes,
-      }),  
+      }),
+      uno({
+        presets: [presetWarp()],
+      }),
       // litElementTailwindPlugin({ mode }),
       mode !== 'lib' && createHtmlPlugin({
         minify: false,
--- a/packages/card/index.js
+++ b/packages/card/index.js
@@ -57,7 +57,7 @@ class WarpCard extends LitElement {
   }
 
   get uuButton() {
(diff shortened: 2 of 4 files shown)


vite.config.js
```diff
@@ -77,7 +77,10 @@ export default ({ mode }) => {
         presets: [presetWarp()],
         mode: 'shadow-dom',
         safelist: classes,
-      }),  
+      }),
+      uno({
+        presets: [presetWarp()],
+      }),
```
siljec, line 83: Is this to apply uno when we are not in shadow dom?

BalbinaK: Good question! That is correct. Our web components presented in the preview page have uno CSS in the scope of their shadow DOM only, so that has no effect on slotted elements that come from the outside. That's why I had to add UnoCSS to the preview page in `pages/includes/scripts.js`. Here for example I would need to have CSS for `.my-12` on the page level in order for w-card to show a correctly styled paragraph slot:
`<w-card><p class="my-12">hello</p></w-card>`

We used a similar setup in [@warp-ds/css-docs](https://github.com/warp-ds/css-docs/blob/master/docs/.vitepress/config.js#L305-L319).
Let me know if any of the above isn't clear :)

siljec (approved):

Issue #30, opened 2023-06-29
SSR Friendly Elements


digitalsadhu:
Currently, Warp custom elements are not SSR friendly because references to lit are not externalised which pulls in references to window which break on the server side.

I think it would be provide a version of the components that can be imported and bundled by apps. At the very least this means Lit must be externalised but it would be a good idea to slim down styles to just the styles necessary for each component. At least so long as we continue to use the shadow dom. If the decision is made at some point to not use shadow dom for the components then a different approach should be taken.

For now, I believe we should take the following approach:
1. Publish a bundle for each component separately with:
  a. lit externalised
  b. component classes scoped "safelist" configured to just the classes for that component 
2. Export each component in package.json exports

I've done some experimentation with the repo today and have a proposed approach to consider (I have a basic working prototype locally):

### Proposed Solution

1. Each component gets its own vite.config.js file which looks something like this:

```js
/* eslint-env node */
import { presetWarp } from "@warp-ds/uno";
import uno from "unocss/vite";
import { defineConfig } from "vite";
import { button } from "@warp-ds/component-classes";

const safelist = Object.values(button)
    .map((e) => {
        if (typeof e === 'object') return Object.values(e).map((e) => e.split(/\s/));
        return e.split(/\s/);
    })
    .flat(Infinity);

export default () => {
  return defineConfig({
    plugins: [
      uno({
        presets: [presetWarp()],
        mode: "shadow-dom",
        safelist,
      }),
    ],
    build: {
      emptyOutDir: false,
      outDir: new URL("../../dist/components", import.meta.url).pathname,
        rollupOptions: {
          output: {
            entryFileNames: "button.js",
          },
          input: new URL("./index.js", import.meta.url).pathname,
          external: ["lit", "lit-html", "lit-element", "lit/directives/class-map.js"],
        },
    },
  });
};
```

2. Include this build when publishing

It sucks, but Vite forces you to run separate commands for each build since if you try to just defined multiple entry points it will extract out/code split out common code which we don't want here.

```
npx vite build --config ./packages/button/vite.config.js
npx vite build --config ./packages/alert/vite.config.js
...etc
```

3. Define exports in package.json

```json5
{
  "exports": {
    // existing exports
    ".": "./dist/index.js",
    "./toast": "./dist/api.js",

    // new component exports
    "./components/affix.js": "./dist/components/affix.js",
    "./components/alert.js": "./dist/components/alert.js",
    "./components/attention.js": "./dist/components/attention.js",
[...]

PR #38, opened 2023-08-13
fix(@warp-ds/toast): remove window reference from toast api bundle, fixing server-side rendering

vsandvold:
The toast API relies on a utility function to prevent browser APIs from being called in a non-browser environment (eg. Node). The utility function identifies non-browser environments by checking  if the window object is defined. The function was imported from a package that also imported the classmap function from lit-html, which in itself contains a reference to the window object on top-level of the module. This causes server-side rendering to fail as soon as the imported modules are parsed and evaluated.

To fix this issue, the utility function in question is moved to a separate module that does not contain transitive references to the window object, and it can now safely be used for server-side rendering.

### DIFF
--- a/packages/broadcast/component.js
+++ b/packages/broadcast/component.js
@@ -1,6 +1,6 @@
 import { LitElement, html, css } from 'lit';
 import { repeat } from 'lit/directives/repeat.js';
-import { windowExists } from '../utils';
+import { windowExists } from '../utils/window-exists';
 
 export class WarpBroadcast extends LitElement {
   static properties = {
--- a/packages/toast/api.js
+++ b/packages/toast/api.js
@@ -1,4 +1,4 @@
-import { windowExists } from '../utils';
+import { windowExists } from '../utils/window-exists';
 
 /**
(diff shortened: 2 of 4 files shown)


trygve-lie:
See also: https://github.com/warp-ds/elements/issues/30

trygve-lie:
This can also be checked with the `isServer` method in lit.

When that is said; to improve style loading all web components must extend a super class which does heavy loading of styles from Warp. In this super class we have a similar check (we're not able to depend on the `isServer` method in lit here) so we could expose this check on the super class then components use this one instead.

See: https://github.com/warp-ds/labs/blob/main/warp-element/src/utils.js#L3-L5

vsandvold:
@trygve-lie Sounds like a solid, long-term solution. In the meantime, would you consider publishing this fix (or a similar fix) for the toast api? People are busy migrating to Warp, and a short-term solution is highly appreciated.

siljec (approved):

siljec:
@vsandvold yes, let's merge this as a short term solution :)

PR #47, opened 2023-08-31
chore(vite.config.js): add button shortcuts to uno config and remove …

siljec:
…unused method

Before:
image

Now:
image

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -6,16 +6,10 @@ import { createHtmlPlugin } from 'vite-plugin-html';
 import path from 'path';
 import glob from 'glob';
 import { classes } from '@warp-ds/css/component-classes/classes';
+import { buttons } from '@warp-ds/css/component-classes/shortcuts';
+
 import { MinifyWarpLib } from './.minifier-plugin.js'
 
-let reset;
-async function getReset() {
-  if (reset) return reset;
-  else {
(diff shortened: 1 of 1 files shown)


BalbinaK (approved):

BalbinaK:
@siljec Just a thought - would it be possible to replace those examples that rely on button shortcuts with Button components? We probably should avoid using those shortcuts as they might get deprecated in a near future.

siljec:
Ah, for sure! I didn't know we were planning on deprecating them? I thought some teams needed them? Should I still merge this PR @BalbinaK ?

BalbinaK:
> Ah, for sure! I didn't know we were planning on deprecating them? I thought some teams needed them? 

I think those teams that can't use our Button component had issues with SSR, which may be fixed in the nearest future. Then I thought it would make sense for us to deprecate those classes and encourage the use of our components instead.

> Should I still merge this PR @BalbinaK ?
I think it would actually be nice to use our components instead. Who knows, maybe we use this documentation as a playground somewhere, and then we'd give good example on how to build things in Warp. Wdyt?

siljec:
> I think it would actually be nice to use our components instead. Who knows, maybe we use this documentation as a playground somewhere, and then we'd give good example on how to build things in Warp. Wdyt?

Okay, sounds good, I'll close this PR then :D

PR #48, opened 2023-08-31
fix: use new button classes and improve conditions

AnnaRybkina:
Can be tested here: http://localhost:5178/pages/components/button.html

Before|After
---|---
image|image


BalbinaK (approved):
🚀

PR #51, opened 2023-09-01
fix: remove classes for disabled buttons as there is no such thing in elements

AnnaRybkina:
We can remove conditions for checking if the button is disabled, and therefor also remove usages of the corresponding classes. There is no button in elements that is disabled.


magnuh (approved):

BalbinaK:
@AnnaRybkina except that there was no `disabled` property on this component, was it a design decision to not have a disabled button? If so, does it mean we should remove disabled state from React Button as well?

PR #43, opened 2023-08-28
fix: add sidebar to "local development pages"

siljec:
Add new Lit component "Sidebar Element" to be used on all pages (local development), to make it easier to navigate between components.

PR #54, opened 2023-09-07
fix: add fabrics old template to style our internal docs

AnnaRybkina:
Startpage 👇 
image

Button component 👇 
image


BalbinaK:
Closing as there seems to be no way for us to remove the Fabric header. We can go with a sidebar quick fix from https://github.com/warp-ds/elements/pull/43  for now.

PR #56, opened 2023-09-19
fix: Add fullwidth button prop

felicia-haggqvist:
This ticket adds a fullwidth prop to the button component.

To test: Link with this PR in the css-repo: https://github.com/warp-ds/css/pull/51

### DIFF
--- a/packages/button/index.js
+++ b/packages/button/index.js
@@ -66,9 +66,7 @@ class WarpButton extends kebabCaseAttributes(LitElement) {
     const utility = this.variant === 'utility';
     const pill = this.variant === 'pill';
     const link = this.variant === 'link';
-    console.log("fullwidth: ", this.fullwidth);
 
-    console.log("ccButton.fullWidth: ", ccButton.fullWidth);
     return classNames(
       {
         [ccButton.secondary]: secondary && !this.small && !this.quiet && !this.loading,
--- a/packages/button/index.js
+++ b/packages/button/index.js
@@ -113,6 +113,7 @@ class WarpButton extends kebabCaseAttributes(LitElement) {
         [ccButton.linkSmall]: link && this.small,
         [ccButton.linkAsButton]: !!this.href,
         [ccButton.fullWidth]: this.fullwidth,
+        [ccButton.contentWidth]: !this.fullwidth,
       },
     );
   }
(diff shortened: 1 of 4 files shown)


packages/button/index.js
```diff
@@ -26,7 +26,7 @@ class WarpButton extends kebabCaseAttributes(LitElement) {
     href: { type: String, reflect: true },
     target: { type: String, reflect: true },
     rel: { type: String, reflect: true },
-    buttonClass: { type: String, reflect: true },
```
BalbinaK, line 29: We have copied this component from Fabric DS and there I see we used to add `this.buttonClass` to Button's classes ([see here](https://github.com/fabric-ds/elements/blob/next/packages/button/index.js#L69)). I see that we removed `this.buttonClass` from the list in [this commit](https://github.com/warp-ds/elements/commit/0aae5004b7bb994610dc8ffb6afa8b274d6a1682). I have a feeling it was done by mistake but correct me if I'm wrong @AnnaRybkina. 
I checked that passing `w-full max-w-full` or `w-full! max-w-full!` to that prop would not fix the full-width issue anyway (`max-w-max` always takes precedence), but maybe there are other scenarios where overriding/adding classes to button from the outside would be needed?

packages/button/index.js
```diff
@@ -26,7 +26,7 @@ class WarpButton extends kebabCaseAttributes(LitElement) {
...
+    fullwidth : { type: Boolean, reflect: true },
```
BalbinaK, line 29: I see that in other web components in this package two-word properties would be written in camelCase, and then linked to corresponding attributes in kebab-case (see a `kebabCaseAttributes` wrapper in line 16 or [/packages/epandable/index.js](https://github.com/warp-ds/elements/blob/next/packages/expandable/index.js#L10-L23) for more examples).

BalbinaK (commented):
Nice catch! I had two comments that I thought would be good to mention here 😊

packages/button/index.js
felicia-haggqvist: Nice catch! I will change the prop to "fullWidth" instead. :)

packages/button/index.js
AnnaRybkina: Yea I deleted it by mistake but it is not even in the docs so I haven't thought about it. It doesn't work with unocss as unocss on bulid time doesn't know what it is provided in `buttonClass` and if consumers won't penetrate shadow dom with the styles for the classes then we won't get any styling for anything provided in `buttonClass`. Hopefully this will be fixed by what Web Wlatform team is working on with the constructed stylesheets.
But yes, I agree let's put it back as before. Good point.

BalbinaK (approved):

PR #69, opened 2023-10-06
chore(workflows): set up synchronization with crowdin

BalbinaK:
Implemented the same way as it was done in Vue: https://github.com/warp-ds/vue/pull/65/files


AnnaRybkina (approved):
nice work, let's go 🚀

PR #82, opened 2023-10-18
fix: a11y issues for attention component

felicia-haggqvist:
- Refactor role on the target element so that it now sets "tooltip" if tooltip is active, otherwise it sets it to "img
- Add aria-label to the target element
- Create dynamic functions to handle translation + set different default ariaLabels depending on the type of attention + which direction. 
- Remove aria-label for arrow, since we now have aria-label on the target element of the attention-component


packages/attention/index.js
```diff
@@ -108,11 +114,80 @@ class WarpAttention extends kebabCaseAttributes(LitElement) {
...
+  }
+
   setAriaLabels() {
     if (this._targetEl && !this._targetEl.getAttribute('aria-describedby')) {
       const attentionMessageId = this._messageEl.id || (this._messageEl.id = generateRandomId());
-      this._messageEl.setAttribute('role', 'tooltip');
+      const role = 
+      this._messageEl.setAttribute('role', this.tooltip ? 'tooltip' : 'img');
       this._targetEl.setAttribute('aria-describedby', attentionMessageId);
+      this._messageEl.setAttribute('aria-label', this.defaultAriaLabel())
```
BalbinaK, line 190: I think we will need to accept user's aria-labels and roles in case they want to override our default setting. For that to work we might need to add 2 new properties to this web component (see [Affix props example](https://github.com/warp-ds/elements/blob/next/packages/affix/index.js#L13) for inspiration 😄), and set our defaults only if those properties are `undefined` (empty string should be accepted).

packages/attention/index.js
```diff
@@ -108,11 +114,80 @@ class WarpAttention extends kebabCaseAttributes(LitElement) {
...
+      case 'bottom':
+        return i18n._({
+          id: 'attention.direction.bottom',
+          message: 'pointing down',
+          comment:
+            'Default screenreader message for bottom direction in the attention component',
+        })
+      case 'left':
+        return i18n._({
+          id: 'attention.direction.left',
```
BalbinaK, line 142: Wdyt of `attention.pointingLeft`?

pages/components/attention.html
```diff
@@ -170,25 +170,25 @@ <h2 class="mt-24 mb-16">Accessibility</h2>
...
           Attention element handles accessibility automatically by wrapping its slotted content with
-          a <code>div</code> with <code>role="tooltip"</code>, and setting an
-          <code>aria-describedby</code> attribute on the target element.
+          a <code>div</code> with either <code>role="tooltip"</code> for tooltip or <code>role="img"</code> for callout and popover, a default <code>aria-label</code>and setting an
+          <code>aria-details</code> attribute on the target element. The default <code>aria-label</code> also supports i18n.
         </p>
         <p>
           It is possible to tell assistive technologies to recognize only a part of Attention's text
-          content. To do that set <code>role="tooltip"</code> on the relevant text element nested in
+          content. To do that set the <code>role</code> and <code>aria-label</code> attributes on the relevant text element nested in
```
BalbinaK, line 178: You don't need to set `aria-label` on the referenced text element. As long as it has a `role`, screen reader will read its contents (text).

pages/components/attention.html
```diff
@@ -170,25 +170,25 @@ <h2 class="mt-24 mb-16">Accessibility</h2>
...
-          <code>aria-describedby</code>. The <code>aria-describedby</code> attribute is on the
+          <code>aria-details</code>. The <code>aria-details</code> attribute is on the
           target element, not on <code>w-attention</code>.
         </p>
 
         <syntax-highlight>
           <w-attention placement="top" popover>
             <div slot="message">
-              <p id="aria-content" role="tooltip">I'm a popover with ARIA "tooltip" role</p>
+              <p id="aria-content" role="tooltip" aria-label="popover with ARIA 'tooltip' role">I'm a popover with ARIA "tooltip" role and a aria-label</p>
```
BalbinaK, line 187: Same here, role="tooltip" will tell screen readers that it should read whatever text this paragraph holds. Does that make sense?

BalbinaK (commented):
We're almost there 🎉

packages/attention/index.js
```diff
@@ -102,39 +112,111 @@ class WarpAttention extends kebabCaseAttributes(LitElement) {
...
   }
 
+  activeAttentionType() {
+    switch (true) {
+      case this.tooltip:
+        return i18n._({
+          id: 'attention.tooltip',
+          message: 'tooltip speech bubble',
+          comment:
+            'Default screenreader message for tooltip speech bubble in the attention component',
```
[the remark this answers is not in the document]
felicia-haggqvist: Ah ok, I will remove it for tooltip! :)

pages/components/attention.html
felicia-haggqvist: Ah goood to know! 😄 I'll remove it then :)

BalbinaK (approved):

PR #86, opened 2023-10-23
fix: Add role property to Box component

magnuh:


felicia-haggqvist (approved):

packages/box/index.js
```diff
@@ -35,9 +36,13 @@ class WarpBox extends LitElement {
     });
   }
 
+  get _optOutRoleWithDefault() {
+    return this.role === '' ? nothing : this.role ?? 'region'
```
BalbinaK, line 40: I'm not familiar with `nothing`, but if we want to make sure that `role` attribute is hidden when `''` is assigned to it, we could use `ifDefined(this._optOutRoleWithDefault)` directive in line 45 ([lit docs](https://lit.dev/docs/templates/directives/#ifdefined)).

magnuh: https://lit.dev/docs/v1/lit-html/writing-templates/#rendering-nothing

BalbinaK: Oh, that's cool! Thanks!

magnuh: It gets a bit "messy" when we want a fallback AND still be able to remove the attribute. 😅

PR #87, opened 2023-10-23
feat: SSR safe individual elements

imprashast:


AnnaRybkina (approved):
Tested in my test app. I screenshot:ed bundle size of the index.js(which is a bundled app in my case) before and after this change in this PR. You can also see that in the Before screenshot elements from node_modules take most of the size, and it is different and more logical on the After screenshot. People were complaining about huge increase of the bundle size when importing elemens as a whole. It can be seen pretty clear on the Before screenshot.

Before|After
---|---
image|image

Great work, components are working as well. Tested Button and Alert. HAVEN'T tested Breadcrumbs though

PR #94, opened 2023-10-27
fix: let classes be applied from outside

imprashast:


packages/button/index.js
```diff
@@ -35,7 +35,7 @@ class WarpButton extends kebabCaseAttributes(WarpElement) {
     target: { type: String, reflect: true },
     rel: { type: String, reflect: true },
     fullWidth: { type: Boolean, reflect: true },
-    buttonClass: { type: String, reflect: true },
+    class: { type: String, reflect: true },
```
magnuh, line 38: Naming this simply `class` will add it both to the `<w-button>` element outside of shadow-dom AND to the wrapping element inside shadow-dom? Right? 🤔

magnuh: ...and to me it seems like the outer elements gets the styling from `text-l`, but inside shadow-dom that class is not defined? 🤷

PR #103, opened 2023-11-02
fix: a11y expandable

felicia-haggqvist:
- Replace use of @warp-ds/icons to inline-svg for expandable-component due to issues with setting css-classes on an element inside the shadow DOM.
- Handle translations of svg-titles directly in the expandable component (Note: uses the same translation-id as for chevron-down and chevron-up icons in @warp-ds/icons to avoid duplicated translations in CrowdIn).

- Switch between chevron-down and chevron-up icons in order to get the correct svg-title (but still keeping the animation).


pages/includes/scripts.js
```diff
@@ -2,7 +2,6 @@ import '../../index.js';
 import 'uno.css';
 import { toast, updateToast, removeToast } from '../../index.js';
 import { WarpToastContainer } from '../../packages/toast/toast-container.js';
-import '@warp-ds/icons/elements';
```
imprashast, line 5: Are icons used in other components? If yes, we'll need to add individual imports for those or else those icons won't work.

imprashast (approved):

packages/expandable/index.js
```diff
@@ -48,28 +66,61 @@ class WarpExpandable extends kebabCaseAttributes(WarpElement) {
...
+  }
+
+  get _chevronDownClasses() {
+    return fclasses({
+      [ccExpandable.chevronTransform]: true,
+      [ccExpandable.chevronExpand]: this.expanded && !this._showChevronUp,
+    })
+  }
+
+  chevronDownTitle() {
```
BalbinaK, line 104: Was there a reason you didn’t use a “get” like with _chevronDownClasses? Also, if this is a private accessor, I suppose we should try to use “_” like suggested in the[ Lit docs](https://lit.dev/docs/components/properties/#:~:text=we%20also%20recommend%20using%20a%20convention%20like%20a%20leading%20underscore%20(_)%20to%20identify%20private%20or%20protected%20properties%20for%20javascript%20users.).

felicia-haggqvist: Ah you're right. I should change it to use an underscore then :)

PR #107, opened 2023-11-08
fix: Add part to the expandable to be able to style an svg inside the shadow-DOM

felicia-haggqvist:
**Note:** Before this PR can be merged, we need to update the version of @warp-ds/icons once [this](https://github.com/warp-ds/icons/pull/50) PR is merged, as well as, update the version of @warp-ds/drive once [this](https://github.com/warp-ds/drive/pull/170) PR is merged.

- Replace the inline-svgs with `chevron-down` and `chevron-up` icons from `@warp-ds/icons`
- Add `part` ([CSS pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/::part)) for each css-class in the `_chevronUpClasses` and `_chevronDownClasses`


packages/expandable/index.js
```diff
@@ -65,6 +61,8 @@ class WarpExpandable extends kebabCaseAttributes(WarpElement) {
       ::slotted(:last-child) {
         margin-bottom: 0px !important;
       }
+
+      @unocss-placeholder;
```
BalbinaK, line 65: It could be this is fine, but I thought that we should keep all classes used in our components in the [@warp-ds/css/component-classes/index.js](https://github.com/warp-ds/css/blob/next/component-classes/index.js), so that  all necessary CSS is available through WarpElements styles - in this case through `components.css` stylesheet. 
Have you checked by linking this repo if this component can be used without issues in a project that is on old @warp-ds/uno version?

vite.config.js
```diff
@@ -70,6 +71,11 @@ export default ({ mode }) => {
       mode !== 'lib' && uno({
         presets: [presetWarp()],
       }),
+      mode === 'development' && uno({
```
imprashast, line 74: Wont you need these classes when running in lib mode?

packages/expandable/index.js
```diff
@@ -65,7 +61,8 @@ class WarpExpandable extends kebabCaseAttributes(WarpElement) {
       ::slotted(:last-child) {
         margin-bottom: 0px !important;
       }
-    `,
+      @unocss-placeholder
```
imprashast, line 64: Based on your comment `Add an additional uno-setup for development-mode in the vite-config.js-file that includes mode: 'shadow-dom' and safeList component-classes, making it possible to test component-classes locally in elements-repo.`, since this is only needed for debug purpose, maybe we don't need to add it to the bundle?

imprashast (approved):
Left 2 comment, but apart from that LGTM 👍

vite.config.js
felicia-haggqvist: are you referring to the "safeList: classes" that I've added for the development mode, and that it should also be in the lib-mode?

BalbinaK: > Wont you need these classes when running in lib mode?

We checked that when you run `pnpm dev`, the mode is 'development'. The goal with this additional uno setup was to test changes from the @warp-ds/css repo locally without too much effort (WarpElement superclass made it a bit tricky).
So if you have some classes that have changed in @warp-ds/css and you want to test them in @warp-ds/elements before making any releases, you could link that local repo here using `npm link`, add `@unocss-placeholder` to styles of the component you're testing, and this way you can see your new styles when running `pnpm dev`. We just need to remember to remove the `@unocss-placeholder` when we're done testing. I hope that makes sense 😊

felicia-haggqvist: I have now updated the ReadMe-file so that it is clearer on the use of `@unocss-placeholder` and to inform that we need to remove it once we're done testing in this commit: [eb79307](https://github.com/warp-ds/elements/pull/107/commits/eb793076d71a7da601bf67cae4cb13dccccd0bdd) :)

README.md
```diff
@@ -47,17 +48,28 @@ import '@warp-ds/elements/components/toast';
...
+
+```js
+static styles = [
+    css`
+      @unocss-placeholder
+      `,
+  ]
+```
+
+Once you have tested it locally, make sure to remove the `@unocss-placeholder` from the static styles before pushing any changes to the repository.
```
BalbinaK, line 72: Sounds great! Would it make sense to suggest to run `pnpm dev` before the `Once you have tested...`? Something like:
```
Run `pnpm dev` to see the linked styles in action.
```

BalbinaK (approved):

README.md
felicia-haggqvist: Good point! I’ll add that :)

BalbinaK (approved):

PR #110, opened 2023-11-15
Fix breadcrumbs for SSR

imprashast:
There are still some ToDo's:

1. Add 2 extra classes in the css tokens for separator
2. Add `inline-flex items-center` in breadcrumb.item


felicia-haggqvist:
The build for i18n-check has failed. When checking the logs, it looks like, for some reason, it doesn't recognize lingui when trying to run `pnpm messages:compile`

PR #111, opened 2023-11-16
fix: eik publishing for elements

felicia-haggqvist:
Solves this JIRA-ticket: [WARP-381](https://nmp-jira.atlassian.net/browse/WARP-381)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ import from '@warp-ds/elements'
 ```
 or a specific component: 
 ```js
-import '@warp-ds/elements/packages/toast'
+import '@warp-ds/elements/components/toast'
 ```
 
 ### Documentation
--- a/README.md
+++ b/README.md
@@ -35,11 +35,11 @@ pnpm add @warp-ds/elements
 ### Import components
 This project builds both the whole Elements package and separate component packages that are published to EIK, making it possible for the user to choose to either import all of the components:
 ```js
-import from '@warp-ds/elements'
+import '@warp-ds/elements';
 ```
 or a specific component: 
 ```js
-import '@warp-ds/elements/components/toast'
+import '@warp-ds/elements/components/toast';
 ```
 
 ### Documentation
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ The project is also using a separate esbuild config for building assets for Eik 
 mapping to ECMAScript modules.
 
 Furthermore, it builds both the whole Elements package and separate component packages that are published to EIK, making it possible for the user to choose to either import all of the components or just a specific one.
+The component packages will also be included in the import map-list for [custom elements](https://github.schibsted.io/nmp-web/eik-import-maps/blob/main/maps/custom-elements/map.json).
 
 
 ## Changelog
--- a/README.md
+++ b/README.md
@@ -32,6 +32,15 @@ pnpm add @warp-ds/elements
 ```html
 <script src="https://assets.finn.no/pkg/@warp-ds/elements/v1"></script>
 ```
+#### Import components
+This project builds both the whole Elements package and separate component packages that are published to EIK, making it possible for the user to choose to either import all of the components:
+```js
+import from '@warp-ds/elements'
+```
+or a specific component: 
+```js
+import '@warp-ds/elements/packages/toast'
+```
 
 ### Documentation
 
@@ -43,12 +52,6 @@ see the [Warp Design System documentation](https://warp-ds.github.io/tech-docs/)
 
 This project is continuously published to [NPM](https://www.npmjs.com/package/@warp-ds/elements) and [Eik](https://assets.finn.no/pkg/@warp-ds/elements) using a `next` tag (e.g. `1.1.0-next.1`).
 Anyone needing to use the latest changes of this package can point to the `next` version while waiting for the stable release.
-
-The project is also using a separate esbuild config for building assets for Eik and includes [Eik's Esbuild plugin](https://github.com/eik-lib/esbuild-plugin), which imports maps for `lit` from an Eik server, as well as, applying
-mapping to ECMAScript modules.
-
-Furthermore, it builds both the whole Elements package and separate component packages that are published to EIK, making it possible for the user to choose to either import all of the components or just a specific one.
-The component packages will also be included in the import map-list for [custom elements](https://github.schibsted.io/nmp-web/eik-import-maps/blob/main/maps/custom-elements/map.json).
 
 
 ## Changelog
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -104,7 +106,13 @@
   },
   "eik": {
     "server": "https://assets.finn.no",
-    "files": "dist"
+    "type": "package",
+    "files": "eik",
+    "import-map": [
+      "https://assets.finn.no/map/react/v2",
+      "https://assets.finn.no/map/vue/v1",
```
BalbinaK, line 113: It doesn't look like it's relevant for this repo :)

package.json
```diff
@@ -104,7 +106,13 @@
...
+      "https://assets.finn.no/map/custom-elements/v2"
```
BalbinaK, line 114: Are we sure about this url path? Where does it come from? I'm mostly curious about the `v2` part of it.

build-elements-eik.js
```diff
@@ -0,0 +1,59 @@
...
+const { glob } = pkg
+
+await eik.load()
+const imports = glob.sync('packages/**/index.js')
+
+imports.forEach(async (item) => {
+  const regex = /\/(\w+)\//
+  const match = item.match(regex)
+
+  if (item.includes('utils')) return
```
BalbinaK, line 13: Not sure it's relevant, but should we also export `unstyled-heading` and `expand-transition` that's part of `packages/utils/`? In the [./index.js](https://github.com/warp-ds/elements/blob/next/index.js#L16) file we do export them.

package.json
imprashast: Yes, it's from here:
https://github.schibsted.io/nmp-web/eik-import-maps/blob/main/maps/custom-elements/eik.json

Not sure why they moved to v2, but that's the latest :)

package.json
imprashast: True

build-elements-eik.js
imprashast: No one uses these outside the elements repo, hence we are not building them. I think utils are meant for internal use in the repo. But correct me if I'm assuming wrong here :)

package.json
BalbinaK: Those paths are the same as the one I found in @fabric-ds/icons: https://github.com/fabric-ds/icons/blob/main/package.json#L69-L72
It doesn't sound correct to override anything on those paths 🤔

build-elements-eik.js
BalbinaK: Makes sense to treat them as internal components unless someone requests access to them, but at the same time our NPM package and the complete package you publish to Eik includes them. I also found one example of expand-transition usage from Fabric (https://finn-codesearch.svc.prod.finn.no/search?q=expand-transition&num=50). I don't mind either approach. I only think it would be good to align on what we make available in Eik and NPM, if that makes sense 😊 

On a side note, it would help review a PR if decisions like this were mentioned in the PR description. It is apparent when one looks deeper into the repo, but with so much going on it can easily be missed 😊

package.json
BalbinaK: Ah, nevermind that, that's the version of maps/custom-elements, which hold import maps for different packages, including @fabric-ds and @warp-ds. https://github.schibsted.io/nmp-web/eik-import-maps/blob/main/maps/custom-elements/map.json

BalbinaK (commented):
It was probably mentioned in some meeting, but I still don’t understand something. In the linked ticket the description says 
```
"Currently the version published on EIK for elements
doesn’t work for importing 'lit' and other dependencies". 
```

What I don't understand is how publishing separate components to Eik fixes a problem with importing 'lit'. It looks like we are externalising ‘lit’ in the esbuild config both when building the whole Elements package, and when building separate components.

Also, with the current change, will it be possible to import the components in a different way than so far? If so, could we mention it in the README.md?

BalbinaK (approved):
I now understand that this PR handles 2 things: 
1. fixes the linked issue with `Currently the version published on EIK for elements doesn’t work for importing 'lit' and other dependencies` by building assets for Eik using a separate esbuild config (`build-elements-eik-js`), which includes [Eik's Esbuild plugin](https://github.com/eik-lib/esbuild-plugin). That plugin loads import maps for `lit` from an [Eik server](https://eik.dev/) and applies the mapping to ECMAScript modules in preparation for upload to the same server.
2. Additionally, as a result of a discussion with teams that only need 1 component from Warp Elements, this PR builds separate component packages to be published to the Eik server. Those component packages will also be included in the import map list for custom elements (https://github.schibsted.io/nmp-web/eik-import-maps/blob/main/maps/custom-elements/map.json).

If I understand it correctly, with the second change, users will be able to use our components like this:
```
import '@warp-ds/elements/packages/toast'
```

I think it would be good to update the README so we make it clear in the context of this package :)

felicia-haggqvist:
@BalbinaK Just updated the Readme-file to inform about these changes: https://github.com/warp-ds/elements/pull/111/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 :)

felicia-haggqvist:
> > @BalbinaK Just updated the Readme-file to inform about these changes: https://github.com/warp-ds/elements/pull/111/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 :)
> 
> Cool! I may have been unclear but I was mainly thinking about adding an example of importing separate components to the [Hot to use](https://github.com/warp-ds/elements#how-to-use) section of the README.md 😅

Ah I see! That makes sense, I can update it! 😅

imprashast:
A short typo @BalbinaK :
`'@warp-ds/elements/components/toast'`
and not `'@warp-ds/elements/packages/toast'`

felicia-haggqvist:
@BalbinaK  @imprashast  Updated readme to showing example of how to import the components: https://github.com/warp-ds/elements/pull/111/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

README.md
```diff
@@ -32,6 +32,15 @@ pnpm add @warp-ds/elements
 ```html
 <script src="https://assets.finn.no/pkg/@warp-ds/elements/v1"></script>
 ```
+### Import components
+This project builds both the whole Elements package and separate component packages that are published to EIK, making it possible for the user to choose to either import all of the components:
+```js
+import from '@warp-ds/elements'
```
BalbinaK, line 38: import '@warp-ds/elements';

felicia-haggqvist: whoops! I'll change it 😅 🙈

PR #116, opened 2023-11-23
fix(textfield): allow styling input's left padding when prefix is wider than 40px

BalbinaK:
Fixes [WARP-374](https://nmp-jira.atlassian.net/browse/WARP-374)

Prefix is an absolutely positioned component and the gap between that component and the `input`'s content is handled using a left padding on the `input`. It’s currently hard-coded to 40px, but with this quick fix we can let users change it by setting a CSS variable in `TextField`’s class (`<TextField class=”[--w-prefix-wdith: 100px]”>`). It’s not dynamic and users will need to calculate how much padding their input needs, but it’s the only non-breaking quick fix we managed to come up with. We should really consider rewriting these form components completely to properly approach styling of different text fields, affixes etc.

Screenshot with long prefix
Screenshot with 2-letter prefix
Screenshot with no prefix

### DIFF
--- a/packages/textfield/index.js
+++ b/packages/textfield/index.js
@@ -39,6 +39,9 @@ class WarpTextField extends WarpElement {
       ::slotted(:last-child) {
         margin-bottom: 0px !important;
       }
+      .warp-input-with-prefix {
+        padding-left:var(--w-prefix-width, 40px);
+      }
     `,
   ];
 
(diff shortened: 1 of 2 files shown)


packages/textfield/index.js
```diff
@@ -54,7 +57,9 @@ class WarpTextField extends WarpElement {
       [input.disabled]: this.disabled,
       [input.readOnly]: this.readOnly,
       [input.suffix]: this._hasSuffix,
-      [input.prefix]: this._hasPrefix,
```
BalbinaK, line 57: This class only contained `pl-40`, which sets left padding if a prefix is present. This is now handled in the scoped style of this component, and because of the specificity, which is the same in the case of our utility classes, I decided to remove this line.

imprashast (approved):

PR #126, opened 2024-02-02
feat: Add pill to elements

imprashast:
Adding pill component to elements


felicia-haggqvist:
I get this typeError in the console when running this branch: 
Skärmavbild 2024-02-05 kl  13 08 24

imprashast:
> I get this typeError in the console when running this branch: Skärmavbild 2024-02-05 kl 13 08 24

Good catch. It was this extra import that I didn't remove while copying: https://github.com/warp-ds/elements/pull/126/commits/cc8cd5621b869b70f9add1d289d66ad765c9c4df

felicia-haggqvist (approved):
🚀

packages/pill/index.js
```diff
@@ -0,0 +1,104 @@
...
+  get _closeClasses() {
+    return classNames({
+      [ccPill.button]: true,
+      [this.suggestion ? ccPill.suggestion : ccPill.filter]: true,
+      [ccPill.close]: true,
+    });
+  }
+
+  _onClick() {
+    this.dispatchEvent(new CustomEvent("w-pill-click", { bubbles: true, composed: true }));
```
imprashast, line 59: These were missing. I looked around other Design Systems and they seem to be prepending their events with something, so I went with `w-{componentName}`. Any opinions @BalbinaK @felicia-haggqvist @magnuh @Skadefryd ?

packages/pill/index.js
```diff
@@ -0,0 +1,104 @@
...
+import WarpElement from "@warp-ds/elements-core";
+
+class WarpPill extends WarpElement {
+  static styles = [WarpElement.styles];
+
+  static properties = {
+    canClose: { type: Boolean },
+    suggestion: { type: Boolean },
+    openSRLabel: { type: String },
+    closeSRLabel: { type: String },
```
BalbinaK, line 19: These properties would be bound to lowercase attributes like `canclose` or `opensrlabel`. Whenever we'd have multi-word properties, we would bind camelCase properties with kebab-case attributes using a `kebabCaseAttributes` wrapper on `WarpElement`. If we use that approach here for the sake of consistency, then maybe `openSRLabel` could be renamed so you don't need to pass an attribute like `open-s-r-label="..."`? Wdyt?

packages/pill/index.js
BalbinaK: Sounds good to me. Shoelace library seems to do it this way and they added a section on custom events to the contributing file, which explains their choice: https://github.com/shoelace-style/shoelace/blob/83fe1ff28e88b1982b528161bf8215d61b4f4592/docs/pages/resources/contributing.md?plain=1#L339C1-L343C313

Does it mean we would use the pill component for example like so?
```
<w-pill @w-pill-click="handleClick">...</w-pill>
```

pages/components/pill.html
```diff
@@ -0,0 +1,95 @@
...
+  <%- include('head.html'); -%>
+  <body>
+    <script type="module" src="../sidebar-element.js"></script>
+    <script>
+        document.querySelector('html').addEventListener('w-pill-click', e => {
+            console.log('Click clicked');
+        });
+        document.querySelector('html').addEventListener('w-pill-close', e => {
+            console.log('Close clicked');
+        });
```
BalbinaK, line 11: So if we have multiple pill components, the same message will be logged in the console, regardless which pill is closed. What if we wanted to log a different message for one of the pills? Would it work by passing a `@w-pill-close="console.log('close clicked pill 1')"` on one of them?

imprashast: Ideally everyone would use these events like `<w-pill @w-pill-click="doSomething()">` which is the lit way of doing this. However for the sake of our documentation site, I didn't want to bring in lit just to use these handlers, hence doing it plain js way. 

To define multiple elements specific event handlers, you can tie them to the dom element instead of the whole document :)

packages/pill/index.js
imprashast: Yeah, that sounds good.

packages/pill/index.js
```diff
@@ -0,0 +1,104 @@
...
+          type="button"
+          class="${this._labelClasses}"
+          @click="${this._onClick}"
+        >
+          <span class="${ccPill.a11y}"
+            >${this.openSRLabel
+              ? this.openSRLabel
+              : this.openFilterSrText}</span
+          >
+          <span><slot></slot></span>
```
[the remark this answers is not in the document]
BalbinaK: In that case I don't know if we should have a `span` wrapper around the slot. In the Button component we didn't need that, we'd just have `<slot></slot>` as `button`'s child. That would also open up for slotting more elements inside such button and not just string (provided we're ok with that). Wdyt? 
I'm still unsure if utilizing slot justifies having a different API here than in React and Vue Pill (i.e. no label prop), provided it only should accept a string. But I'm fine with whatever you choose. Just wanted to raise my thoughts.

packages/pill/index.js
BalbinaK: What do you think about `closeSrLabel`/`openSrLabel` so we can have `close-sr-label`/`open-sr-label` attributes?

packages/pill/index.js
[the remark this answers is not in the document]
imprashast: Fair enough :)

packages/pill/index.js
```diff
@@ -0,0 +1,107 @@
...
+    openSRLabel: { type: String },
+    closeSRLabel: { type: String },
+  };
+
+  constructor() {
+    super();
+    activateI18n(enMessages, nbMessages, fiMessages);
+    this.canClose = false;
+    this.suggestion = false;
+
```
BalbinaK, line 29: Should we also initiate `this.label` with `""` here?

imprashast: I think of `label` as a mandatory field here, hence not setting anything

pages/components/pill.html
```diff
@@ -0,0 +1,100 @@
...
+              <td>boolean</td>
+              <td>"false"</td>
+            </tr>
+            <tr>
+              <td>suggestion</td>
+              <td>boolean</td>
+              <td>false</td>
+            </tr>
+            <tr>
+              <td>open-s-r-label</td>
```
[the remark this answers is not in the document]
imprashast: Just aligning all props across React, Vue & Elements, so this should also be the same. Don't you think?

imprashast:
So going back to using the slot and changing to using `open-sr-label` as it makes the most sense in a web-component @BalbinaK

felicia-haggqvist (approved):

PR #127, opened 2024-02-05
feat: add Modal component

pearofducks:
Implementation notes:
- CSS animation needs to be used as opposed to transitions due to browser support. I'm pretty sure Uno supports all the animations this component would use, but Warp doesn't bring those in ATM.
- The `dialog` element is used as a hollow wrapper in this implementation so backdrop-clicks can be distinguished from modal-actual ones.
- I haven't had time to wire up i18n or test if the warp-element icons work instead of the baked in SVGs - I can take care of that but won't have time for a while

TODO-ish notes:
- There's a new set of component-classes that enable the more flexible behavior, as well as aligning the design to the spec in Figma and improving the title-buttons on mobile [1]
- There's a new pill-button variant needed for the bleed image title area, this needs spec'd [1]

[1] - these are blocked by design, Adi has been busy. It makes more sense for this implementation to get taken over by the DS team than me try to work from the outside with getting those unblocked. :D


packages/modal/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+if (!customElements.get('w-modal')) {
+  customElements.define('w-modal', ModalMain)
+  customElements.define('w-modal-title-bar', ModalTitleBar)
+  customElements.define('w-modal-title-close', ModalTitleClose)
+  customElements.define('w-modal-footer', ModalFooter)
+}
+
+export {
+  ModalMain as WarpModal,
```
imprashast, line 15: Maybe there needs to be only 1 export here rather than all these, since you just need to import the file and you'll get all the custom elements thru the `.define` rather than these exports

packages/modal/modal-footer.js
```diff
@@ -0,0 +1,18 @@
...
+  render() {
+    return html`
+      <div class="${cc.footerSlot}">
+        <slot @slotchange="${this.handleSlotChange}"></slot>
+      </div>
+    `
+  }
+  static styles = [
+    WarpElement.styles,
+    css`@unocss-placeholder`
```
imprashast, line 16: The `@unocss-placeholder` won't be needed since all the styles will be in the WarpElement

packages/modal/component-classes.js
```diff
@@ -0,0 +1,50 @@
...
+const buttonTypes = { primary: `border-0 rounded-8 ${buttonDefaultStyling}`, secondary: `border-2 rounded-8 ${buttonDefaultStyling}`, utility: `border rounded-4 ${buttonDefaultStyling}`, negative: `border-0 rounded-8 ${buttonDefaultStyling}`, pill: `p-4 rounded-full border-0 inline-flex items-center justify-center hover:bg-clip-padding ${buttonDefaultStyling}`, link: `bg-transparent focusable ease-in-out inline active:underline hover:underline ${buttonColors.link}`, };
+const buttonTextSizes = { medium: 'text-m leading-[24]', xsmall: 'text-xs', };
+
+// TODO: can hardcode these into the classes below if this spacing is actually ratified
+const mobileBorderPadding = 24
+const borderPadding = 24
+export const modalV2 = {
```
imprashast, line 26: I'll move this to the css repo with the name v2 since all these are different classes.

packages/modal/index.js
pearofducks: Again just following convention from other components 🤷

pearofducks:
The recent set of commits aligns with the improved API the Warp team suggested. I've also added animation for when the 'back' button shows/hides - it was agreed in the Fabric days this was a GoodThing™️ since it helped the user see the change in the titlebar area.

A 'titlebar' modal now looks like this:

```html
<w-modal>
  <w-modal-header slot="header" title="An example modal" ?back="hasBack"></w-modal-header>
  <div slot="content">
    <p>Wu-Tang is forever</p>
  </div>
  <w-modal-footer slot="footer">
    <w-button variant="primary">OK</w-button>
  </w-modal-footer>
</w-modal>```

and an image-on-top modal looks like this:

```html
<w-modal>
  <w-modal-header slot="header" title="Look a doggo!">
    <img slot="top" class="h-[256] w-full object-cover" src="https://source.unsplash.com/random?dog" alt="a picture of a dog">
  </w-modal-header>
  <w-modal-footer slot="footer">
    <w-button variant="primary">OK</w-button>
  </w-modal-footer>
</w-modal>
```

pearofducks:
https://github.com/warp-ds/elements/assets/736895/3c72e056-feed-4875-99e7-4f7b1274a21d

Animation video for reference

PR #144, opened 2024-04-05
chore: reuse animation library from modal in expandable

pearofducks:
Two primary goals in this PR:
- Reuse `@itsy/animate` that the Modal component uses for transitioning the title to ensure small bundle size
- Simplify the `expand-transition` component, might need some checks from @digitalsadhu to ensure I haven't missed something critical - but I don't understand why some of the moving parts are in the original

This PR can safely wait for the Modal's merge, so no rush to review

(I still need to replace the element-collapse use in Toast, but I don't anticipate that being a blocker and wanted to get an RFC first on this change)


pearofducks:
@digitalsadhu - just quickly can you see if there's anything in the `expand-transition` component I'm changing that is obviously bad? (it did a few obscure things I don't understand the rationale for)

BalbinaK:
@pearofducks After running the docs locally on Safari, I noticed that the content doesn't hide/show on expand/collapse the same way it does on Chrome and Firefox. The previous implementation worked fine on Safari, which can be checked [here](https://warp-ds.github.io/elements/expandable.html).
expanded-transition-safari

pearofducks:
@BalbinaK - thanks for looking into this and spotting that quirk, I'll take a look!

PR #147, opened 2024-04-11
fix: Styling fixes for Card component

magnuh:
Utilise styling fixes from component classes in the Card component.

https://github.com/warp-ds/elements/assets/7531505/3c7e230e-453c-41b3-92ae-d3e5d67f9170

### DIFF
--- a/packages/card/index.js
+++ b/packages/card/index.js
@@ -40,21 +40,18 @@ class WarpCard extends WarpElement {
     this.clickable = false;
   }
 
-  get _outerClasses() {
+  get _containerClasses() {
     return fclasses({
       [ccCard.card]: true,
       [ccCard.cardShadow]: !this.flat,
-      [this.selected ? ccCard.cardSelected : ccCard.cardUnselected]: !this.flat,
+      [ccCard.cardSelected]: !this.flat && this.selected,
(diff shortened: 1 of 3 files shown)


felicia-haggqvist (approved):
LGTM! 🎉

packages/card/index.js
```diff
@@ -40,21 +40,18 @@ class WarpCard extends WarpElement {
...
     });
   }
 
-  get _innerClasses() {
-    return fclasses({
-      [ccCard.cardOutline]: true,
-      [this.selected ? ccCard.cardOutlineSelected : ccCard.cardOutlineUnselected]: true,
-    });
+  get _outlineClasses() {
+    return [ccCard.cardOutline, this.selected ? ccCard.cardOutlineSelected : ccCard.cardOutlineUnselected].join(' ');
```
BalbinaK, line 54: Just curious, why change the way you collect these classes? I thought it was to get rid of `fclasses` call, but we still use it in line 44.

magnuh: Basically because it feels like crossing the river for water, when all conditions are `true`. 🤷 Also I don't understand the name. So these combined makes it hard to understand the code. But I reverted this change... and the bigger picture of this can be handled later.

PR #148, opened 2024-04-16
chore: Remove info type leftovers from Toast

magnuh:
Reported here: https://sch-chat.slack.com/archives/C04P0GYTHPV/p1713216068414779


imprashast (approved):
Nice 👍

PR #149, opened 2024-04-22
feat: add more functionality to attention component

felicia-haggqvist:
**Part of fixes for Jira-tickets:** [WARP-432](https://nmp-jira.atlassian.net/browse/WARP-432), [WARP-341](https://nmp-jira.atlassian.net/browse/WARP-341) & [WARP-373](https://nmp-jira.atlassian.net/browse/WARP-373)

- Add `cross-axis` prop
- Only render attention component when `targetEl` is not undefined or when it is a `callout`
- Bump to latest version of `@warp-ds/core` dependency

### DIFF
--- a/packages/attention/index.js
+++ b/packages/attention/index.js
@@ -177,7 +177,7 @@ class WarpAttention extends kebabCaseAttributes(WarpElement) {
   }
 
   get _targetEl() {
-    const targetSlot = this.renderRoot.querySelector("slot[name='target']");
+    const targetSlot = this.renderRoot?.querySelector("slot[name='target']");
     return targetSlot ? targetSlot.assignedNodes()[0] : null;
   }
 
--- a/packages/attention/index.js
+++ b/packages/attention/index.js
@@ -344,7 +344,7 @@ class WarpAttention extends kebabCaseAttributes(WarpElement) {
       this.close();
     }
   }
-  
+
   render() {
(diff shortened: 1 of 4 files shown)


packages/attention/index.js
```diff
@@ -342,27 +346,27 @@ class WarpAttention extends kebabCaseAttributes(WarpElement) {
...
-          : html`
-              <div id="attention" class="${this._wrapperClasses}">
-                <slot name="message"></slot>
-                ${this._arrowHtml} ${this.canClose ? this._closeBtnHtml : nothing}
-              </div>
-              <slot name="target"></slot>
-            `}
-      </div>
-    `;
+    return html` ${this.callout || (this._targetEl !== undefined && !this.callout)
```
BalbinaK, line 349: I suppose this has something to do with changes introduced in the core library (https://github.com/warp-ds/core/pull/16), but I'm not sure what exactly. If this check is necessary, maybe we could turn it around for better readability, like it was done in the case of Toast (https://github.com/warp-ds/elements/blob/next/packages/toast/toast.js#L152)?

felicia-haggqvist: Updated in this commit: [c10e8ce](https://github.com/warp-ds/elements/pull/149/commits/c10e8ce967df3e229b1486aadd9b546a33242fb1) (I fixed the lint issue in this commit 🙈)

BalbinaK (approved):
LGTM!

PR #150, opened 2024-05-03
fix(textfield): clean up classes

BalbinaK:
## Description
Use cleaned up input component classes to avoid buggy style overriding when utility classes styling the same CSS properties were applied to the same element

## Changes included
- Update @warp-ds/uno to 1.12.0 in order to handle padding classes with arbitrary values that contain CSS variables (e.g. `pl-[var(--w-prefix-width,_40px)]`
- Use the cleaned up input component classes to avoid style overriding
- replaced fclasses() with classes() call to avoid `undefined` classes overriding the defined ones
- Set the pnpm version to v9 in github workflows to avoid lockfile issues after updating pnpm locally

## Testing
Tested with TextField and Textarea.
Gif representing TextField in different states behaving and looking the same as before

### DIFF
--- a/packages/textfield/index.js
+++ b/packages/textfield/index.js
@@ -2,7 +2,7 @@ import { css, html } from 'lit';
 import WarpElement from '@warp-ds/elements-core';
 import { input, label as l, helpText as h } from '@warp-ds/css/component-classes';
 import { ifDefined } from 'lit/directives/if-defined.js';
-import { fclasses } from '../utils';
+import { classes } from '../utils';
 
 class WarpTextField extends WarpElement {
   static properties = {
@@ -40,9 +40,6 @@ class WarpTextField extends WarpElement {
       ::slotted(:last-child) {
         margin-bottom: 0px !important;
       }
-      .warp-input-with-prefix {
-        padding-left: var(--w-prefix-width, 40px);
-      }
     `,
   ];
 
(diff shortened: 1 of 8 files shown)


packages/textfield/index.js
```diff
@@ -52,20 +49,19 @@ class WarpTextField extends WarpElement {
...
 
   get _inputStyles() {
-    return fclasses({
-      [input.default]: true,
-      [input.invalid]: this.invalid,
-      [input.disabled]: this.disabled,
-      [input.readOnly]: this.readOnly,
+    return classes({
+      [input.base]: true,
+      [input.default]: !this.invalid && !this.disabled && !this.readOnly,
```
BalbinaK, line 54: The way `fclasses` util was handling arguments made that classes strings with `undefined` value would override classes with truthy value. Here e.g. `'border-1': true` (`[input.default]: !this.invalid ...`) would be overridden by `border-1: undefined` (`[input.default]: this.invalid ...`). The resulting style would be like this:
Screenshot 2024-05-03 at 12 32 34

packages/textfield/index.js
```diff
@@ -52,20 +49,19 @@ class WarpTextField extends WarpElement {
...
+      [input.invalid]: this.invalid && !this.disabled && !this.readOnly,
+      [input.disabled]: !this.invalid && this.disabled && !this.readOnly,
+      [input.readOnly]: !this.invalid && !this.disabled && this.readOnly,
       [input.suffix]: this._hasSuffix,
-      // we style input with prefix here because we cannot use
-      // arbitrary values with commas in UnoCSS like pl-[var(--w-prefix-width, 40px)]
-      'warp-input-with-prefix': this._hasPrefix,
+      [input.prefix]: this._hasPrefix,
```
felicia-haggqvist, line 59: Right now the prefix ends up right on top of the placeholder/input value: 
Skärmavbild 2024-05-06 kl  16 01 59
Skärmavbild 2024-05-06 kl  16 02 04

BalbinaK: Sounds like you might be missing the`@unocss-placeholder` in textfield component's static styles 🤔

felicia-haggqvist: But is that something I should add myself? Isn't that something that should already be in the textfield component in this branch? 🤔

BalbinaK: I can understand how this can be confusing! In this package, we are relying on WarpElement's styles, which are always coming from Eik (https://assets.finn.no/pkg/@warp-ds/css/v1). Until the updated styles are released from @warp-ds/css and the v1 alias is updated in Eik, the only way to test our css changes in this repo is by generating styles with UnoCSS locally. For that to work inside Shadow DOM, we need to add a temporary `@unocss-placeholder` to the static styles of a given web component, so UnoCSS understands where to inject the generated CSS. It was explained in our README.md in the section ["How to test component-classes locally"](https://github.com/warp-ds/elements?tab=readme-ov-file#how-to-test-component-classes-locally) 😊

felicia-haggqvist: Ah right! I forget this all the time! 😄 🙈

felicia-haggqvist: Yeah now it works like it should! 😄 👍

felicia-haggqvist (approved):
LGTM! 🎉

felicia-haggqvist (approved):

PR #153, opened 2024-05-16
fix: update warp-ds eslint-config devDep and add .eslintrc-file

felicia-haggqvist:
Fixes Jira issue: [WARP-520](https://nmp-jira.atlassian.net/browse/WARP-520)

- Bump version of `@warp-ds/eslint-config` devDep
- Add .eslintrc-file
- lint all code using `pnpm lint`
- Remove `eslint-plugin-prettier` and `eslint-config-prettier` as devDependencies since they are already peerDependencies in @warp-ds/eslint-config

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,9 +63,11 @@
     "element-collapse": "1.1.0",
     "esbuild": "0.20.1",
     "eslint": "8.57.0",
-    "@warp-ds/eslint-config": "1.0.4",
+    "@warp-ds/eslint-config": "1.0.5",
     "eslint-plugin-prettier": "5.1.3",
     "eslint-config-prettier": "9.1.0",
+    "eslint-plugin-import":"^2.29.1",
+    "eslint-plugin-lit":"^1.12.0",
     "express": "4.18.2",
     "glob": "10.3.10",
     "html-format": "1.1.6",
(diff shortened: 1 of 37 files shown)


package.json
```diff
@@ -63,9 +63,8 @@
     "element-collapse": "1.1.0",
     "esbuild": "0.20.1",
     "eslint": "8.57.0",
-    "@warp-ds/eslint-config": "1.0.4",
-    "eslint-plugin-prettier": "5.1.3",
-    "eslint-config-prettier": "9.1.0",
```
BalbinaK, line 68: @imprashast, although all seems to work fine, I'd rather double check with you - if these are defined as `peerDependencies` in @warp-ds/esling-config, does it mean we DO or DO NOT need to install those wherever the package is used?

imprashast: @felicia-haggqvist already did talk to me about these. If you have peerDeps in 1 package, then you can skip it from the one that's using that package :)

imprashast (approved):

package.json
BalbinaK: I've been trying to google around before and my impression was that `peerDependencies` are used when the user of a package like `@warp-ds/eslint-config` is supposed to provide the dependent package. For example this is what I read in this article about peerDependencies: https://dev.to/adamgoth/understanding-package-dependencies-within-a-pnpm-monorepo-19ge

> When you specify a package as a peerDependency in your package.json file, you are essentially saying that "this > package requires that the consumer of my package provides this package".

If I'm misunderstanding some concepts of pnpm/npm/whatever, I would very much appreciate some guidance 😊

PR #158, opened 2024-06-13
fix: merge conflicts, adjust style, cleanup

BalbinaK:
Fixes [WARP-472](https://nmp-jira.atlassian.net/browse/WARP-472)
Dependent on https://github.com/warp-ds/css/pull/214.

Changes:
- ~~replaced WarpElement styles with inlined utility classes + <link href="...resets.css">~~
- fix linting
- updated close button style to align with [Figma sketches](https://www.figma.com/design/oHBCzDdJxHQ6fmFLYWUltf/branch/GInIpbQUuGgFrnDQMBj4BW/WARP---Components-2.0?t=s6LpVtAX4daNlfrW-0)

Me and @Skadefryd wanted to test adding the styles directly in the component and it works fine when building assets with `vite` (developement mode), but we need confirmation that this is a valid setup for production. We wanted the uno-generated css to be inlined in the static styles when the Modal element is published to NPM but we don't see it happening now. Now this setup is using Esbuild to build assets that are then published to NPM/Eik (build.js file), but do we need to include the uno plugin to that setup or use some other build tool for Modal styles to be injected?


pearofducks:
My understanding was that `WarpElement` is providing the base Warp CSS [seen here](https://github.com/warp-ds/labs/blob/main/warp-element/src/utils.js#L62) - and this includes the reset. So if I'm not wrong, the inclusion of the reset via the `link` tag is uncessary for any class that extends `WarpElement`

packages/modal/modal-header.js
```diff
@@ -1,92 +1,111 @@
...
   }
   render() {
     return html`
-      <div class="${cc.titleSlot}">
+      <link href="https://assets.finn.no/pkg/@warp-ds/css/v1/resets.css" rel="stylesheet" />
+      <div class="relative">
         <slot name="top" @slotchange=${this.handleTopSlotChange}></slot>
-        <div class="${!this._hasTopContent && cc.titleBarSlot}">
+        <div
+          class="${this._hasTopContent ? '' : 'pt-16 sm:pt-24 px-16 sm:px-32 grid gap-12 grid-cols-[auto_1fr_auto] items-start shrink-0!'}">
```
pearofducks, line 30: Why a fake-ternary here? Unless this is a stylistic standard established in the Warp team, this is harder (to me) to parse than just using logical-and.

Lit is smart enough IIRC to handle falsey cases and won't put `null` or `false` as a class.

pearofducks:
> we need confirmation that this is a valid setup for production

IIRC - As long as all classes here were in the `components.css` (aka component classes) file that would be fine, but it seems risky.

Reason: most Uno setups won't try to compile classes found in node_modules, so what is working here via Uno catching the styles might not work when published.

packages/modal/modal-header.js
BalbinaK: We added this namely because we saw `class="false"` in the inspected DOM tree locally. So it seemed Lit didn't understand that `"false"` is not a class we want to assign when the left side of `&&` is `false` 😅

pearofducks: You're right! I'm remembering wrong! 🙈

BalbinaK:
> most Uno setups won't try to compile classes found in node_modules, so what is working here via Uno catching the styles might not work when published.

We were hoping that we could replace `@unocss-placeholder` with actual styles upon release to NPM. IIRC, I don't think other UnoCSS setups would try to compile classes if they were inside web components 🤔

pearofducks:
> We were hoping that we could replace `@unocss-placeholder` with actual styles upon release to NPM. IIRC, I don't think other UnoCSS setups would try to compile classes if they were inside web components 🤔

Ahhhh, I'm dumb today, sorry. Yes this should work.

BalbinaK:
As mentioned in the PR description, we wanted to test out using modal classes directly in the web component and relying on UnoCSS to generate styles from there, but we have encountered some issues with the build setup and WarpElement exposing too many stylesheets. This opened a discussion with the platform web team, which showed we need some more testing to approach component styling in an optimal way. Until that is carried out, we decided to revert inlining modal classes in [4658ccc](https://github.com/warp-ds/elements/pull/158/commits/4658ccc962f93bcbcdba219cad9b705a41281f92), so we rely on @warp-ds/css/component-classes after all. This way we should be able to merge this PR sooner than later. 

Before merging, we need to release https://github.com/warp-ds/css/pull/214.

Screenshot of Modal Element with back and close button on mobile screen
Screenshot of Modal Element with back and close button on desktop screen
Screenshot of Modal Element with image and close button on mobile screen
Screenshot of Modal Element with image and close button on desktop screen

pearofducks (approved):

felicia-haggqvist (approved):
LGTM! 🎉

PR #159, opened 2024-06-19
fix(select): clean up classes

felicia-haggqvist:
## Description
This PR ensures no classes styling the same CSS properties are being set on the same HTML element in the Select component.

Additionally this PR has:
-  Added missing props `disabled` and `readOnly` to Select component
- Added missing Select examples for `disabled` and `readOnly`.
- Added a `handleKeyDown()` function to disable `space`, `arrowUp` and `arrowDown` for when `readOnly` prop is set to `true`
- Added a more interactive invalid example (where the invalid state will be removed if the user chooses the right choice). 

## How to test
Either link this branch with @warp-ds/css (`fix/clean-up-select-classes` branch) or replace this for @warp-ds/css dependency in package.json: `github:warp-ds/css#component-classes-cleanup` with this  `github:warp-ds/css#fix/clean-up-select-classes`

### DIFF
--- a/packages/select/index.js
+++ b/packages/select/index.js
@@ -36,29 +36,64 @@ export class WarpSelect extends kebabCaseAttributes(WarpElement) {
     // Whether to show optional text
     optional: { type: Boolean, reflect: true },
 
+    // Renders the field in a disabled state.
+    disabled: { type: Boolean, reflect: true },
+
+    // Renders the field in a readonly state.
+    readOnly: { type: Boolean, relfect: true },
+
     _options: { state: true },
   };
 
   static styles = [WarpElement.styles];
 
+  constructor() {
+    super();
+    activateI18n(enMessages, nbMessages, fiMessages);
+
+    this._options = this.innerHTML;
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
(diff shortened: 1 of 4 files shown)


packages/select/index.js
```diff
@@ -36,29 +36,67 @@ export class WarpSelect extends kebabCaseAttributes(WarpElement) {
...
+
+    super.disconnectedCallback();
+  }
+
+  handleKeyDown(event) {
+    const target = event.target.closest('w-select[data-readonly-id="' + this._readOnlyId + '"]');
+    if (target && (event.key === ' ' || event.key === 'ArrowDown' || event.key === 'ArrowUp')) {
+      event.preventDefault();
+    }
+  }
```
BalbinaK, line 80: Maybe we could utilise the `@keydown` event listener on the `select` element, like it was done in Card and Attention component. Then the `handleKeyDown" method could be more similar to what we have in React Select and the logic would be much less complicated :)

felicia-haggqvist: That makes sense! I'll change it! 😄

felicia-haggqvist: Updated in this commit: [0f706be](https://github.com/warp-ds/elements/pull/159/commits/0f706bee85360bd1ed77633b950f73f9f80f40d0)

BalbinaK (approved):
Looks great!

PR #160, opened 2024-06-25
fix: update release.yml to include retry for semantic-release and only if succesful will it publish to eik

felicia-haggqvist:
Sometimes we get this error when trying to make a semantic release after having merged a PR: 
`"An error occurred while running semantic-release: RequestError [HttpError]: You have exceeded a secondary rate limit. Please wait a few minutes before you try again."`

This PR updates the `release.yml` file so that it will try 3 times to make a semantic release, and only if it has successfully made a semantic-release will it publish to EIK.


BalbinaK (approved):
Let's give it a try!

PR #173, opened 2024-07-30
fix(card): clean up classes

felicia-haggqvist:
## Description
This PR ensures no classes styling the same CSS properties are being set on the same HTML element in the `Card` component.

**Changes:**
- Refactor classes in `Card` component
- Replace `ccCard.card` with `ccCard.base`
- Add translation to replace hardcoded button text
- Only render `<div class="${this._outlineClasses}"></div>` when `this.flat` is false, like we also do in react and vue
- Add another example to showcase how the card looks like when both `flat` and `selected` are true

## How to test
Either link this branch with @warp-ds/css (`fix/cleanup-card-classes` branch) or replace this for @warp-ds/css dependency in package.json: `github:warp-ds/css#component-classes-cleanup` with this  `github:warp-ds/css#fix/cleanup-card-classes`

### DIFF
--- a/packages/card/index.js
+++ b/packages/card/index.js
@@ -75,7 +75,7 @@ class WarpCard extends kebabCaseAttributes(WarpElement) {
       html`<button class="${ccCard.a11y}" aria-pressed="${this.selected}" tabindex="-1">${this.buttonText}</button>`;
     const renderSpan = () => html`<span role="checkbox" aria-checked="true" aria-disabled="true"></span>`;
 
-    return this.clickable ? renderButton() : !this.clickable && this.selected ? renderSpan() : '';
+    return this.clickable ? renderButton() : this.selected ? renderSpan() : '';
   }
 
   keypressed(e) {
@@ -89,7 +89,7 @@ class WarpCard extends kebabCaseAttributes(WarpElement) {
   render() {
     return html`
       <div tabindex=${ifDefined(this.clickable ? '0' : undefined)} class="${this._containerClasses}" @keydown=${this.keypressed}>
-        ${this._interactiveElement} ${!this.flat ? html`<div class="${this._outlineClasses}"></div>` : ''}
+        ${this._interactiveElement} ${this.flat ? '' : html`<div class="${this._outlineClasses}"></div>`}
         <slot></slot>
       </div>
     `;
--- a/packages/card/index.js
+++ b/packages/card/index.js
@@ -1,17 +1,45 @@
 import { html, css } from 'lit';
 
+import { classNames } from '@chbphone55/classnames';
+import { i18n } from '@lingui/core';
 import { card as ccCard } from '@warp-ds/css/component-classes';
 import WarpElement from '@warp-ds/elements-core';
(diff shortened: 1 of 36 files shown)


magnuh (approved):

packages/card/index.js
```diff
@@ -78,8 +88,7 @@ class WarpCard extends WarpElement {
   render() {
     return html`
       <div tabindex=${ifDefined(this.clickable ? '0' : undefined)} class="${this._containerClasses}" @keydown=${this.keypressed}>
-        ${this.clickable ? this.uuButton : ''} ${!this.clickable && this.selected ? this.uuSpan : ''}
-        <div class="${this._outlineClasses}"></div>
+        ${this._interactiveElement} ${!this.flat ? html`<div class="${this._outlineClasses}"></div>` : ''}
```
imprashast, line 91: Wasn't `<div class="${this._outlineClasses}"></div>` added irrespective of if this.flat was true or not? 🤔 

Also this is cleaner I think:
`${this.flat ? '' : html`<div class="${this._outlineClasses}"></div>`}`

packages/card/index.js
```diff
@@ -29,42 +57,24 @@ class WarpCard extends WarpElement {
...
-    return html`<button class="${ccCard.a11y}" aria-pressed="${this.selected}" tabindex="-1">Velg</button>`;
-  }
+  get _interactiveElement() {
+    const renderButton = () =>
+      html`<button class="${ccCard.a11y}" aria-pressed="${this.selected}" tabindex="-1">${this.buttonText}</button>`;
+    const renderSpan = () => html`<span role="checkbox" aria-checked="true" aria-disabled="true"></span>`;
 
-  get uuSpan() {
-    return html`<span role="checkbox" aria-checked="true" aria-disabled="true"></span>`;
+    return this.clickable ? renderButton() : !this.clickable && this.selected ? renderSpan() : '';
```
imprashast, line 77: The `: !this.clickable` should be redundant.

This instead?
`return this.clickable ? renderButton() : this.selected ? renderSpan() : '';`

packages/card/index.js
felicia-haggqvist: In react and vue we only render this div with the outline-classes if the flat prop is false. So I assumed we wanted the same behavior for elements? Otherwise if it should be rendered regardless of whether the flat prop is true or not, then we need to change that in react and vue as well.

felicia-haggqvist: It will look like this if we don't check if this.flat is false: 
Skärmavbild 2024-07-31 kl  14 48 58

and it will look like this if we do check: 
Skärmavbild 2024-07-31 kl  14 49 07

felicia-haggqvist: I do feel however that the outline is more correct in the first example. I mean then the outline is behaving like an outline should behave! :D But we just need to decide on what we want and make sure it looks and behaves the same in react, vue and elements :)

packages/card/index.js
felicia-haggqvist: ah good catch! 😄

imprashast (approved):

PR #174, opened 2024-08-02
fix: update classnames

felicia-haggqvist:
**Changes:**

- Update classnames that have been renamed in component-classes

- Standardize the method of using classNames from the [@chbphone55/classnames](https://www.npmjs.com/package/@chbphone55/classnames) package across all components.
_Before:_ Different components used various methods (arrays, objects) to apply class names, resulting in inconsistent code and potential confusion. 
For example: 
` return fclasses({
      [this._classBase.wrapper]: true,
      [this._classBase.wrapperWithLabel]: this.label,
      [this._classBase.wrapperWithIcon]: !this.label,
    });`
_After:_ All components now consistently use arrays to define class names, which will improve consistency and readability.
For example:
` return classNames([this._classBase.wrapper, this.label ? this._classBase.wrapperWithLabel : this._classBase.wrapperWithIcon]);`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -13,10 +13,10 @@ importers:
         version: 4.11.2
       '@warp-ds/core':
         specifier: 1.1.5
-        version: 1.1.5(@floating-ui/dom@1.6.8)
+        version: 1.1.5(@floating-ui/dom@1.6.10)
       '@warp-ds/css':
         specifier: github:warp-ds/css#component-classes-cleanup
-        version: https://codeload.github.com/warp-ds/css/tar.gz/d54396187a6695970913aa8ff59eeb5bd39b89db(@warp-ds/uno@2.0.0(unocss@0.61.9(postcss@8.4.40)(rollup@4.19.2)(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3))))
+        version: https://codeload.github.com/warp-ds/css/tar.gz/b547413d322cd0245a1145b0c069ddb53fe017f7(@warp-ds/uno@2.0.0(unocss@0.61.9(postcss@8.4.41)(rollup@4.20.0)(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3))))
       '@warp-ds/elements-core':
         specifier: 2.x
         version: 2.0.1
@@ -59,7 +59,7 @@ importers:
         version: 1.0.5(eslint-config-prettier@9.1.0(eslint@8.57.0))(eslint-plugin-import@2.29.1(eslint@8.57.0))(eslint-plugin-prettier@5.2.1(eslint-config-prettier@9.1.0(eslint@8.57.0))(eslint@8.57.0)(prettier@3.3.2))(eslint@8.57.0)
       '@warp-ds/uno':
         specifier: 2.x
-        version: 2.0.0(unocss@0.61.9(postcss@8.4.40)(rollup@4.19.2)(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3)))
+        version: 2.0.0(unocss@0.61.9(postcss@8.4.41)(rollup@4.20.0)(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3)))
       cors:
         specifier: 2.8.5
         version: 2.8.5
@@ -86,7 +86,7 @@ importers:
         version: 1.1.7
       lit:
         specifier: 3.x
-        version: 3.1.4
+        version: 3.2.0
       playwright:
         specifier: 1.45.1
         version: 1.45.1
@@ -101,13 +101,13 @@ importers:
         version: 24.0.0(typescript@5.5.3)
       tap:
         specifier: 21.0.0
-        version: 21.0.0(@swc/core@1.7.4)(@types/node@20.14.10)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(typescript@5.5.3)
+        version: 21.0.0(@swc/core@1.7.6)(@types/node@20.14.10)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(typescript@5.5.3)
       typescript:
         specifier: 5.5.3
         version: 5.5.3
       unocss:
         specifier: 0.x
-        version: 0.61.9(postcss@8.4.40)(rollup@4.19.2)(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3))
+        version: 0.61.9(postcss@8.4.41)(rollup@4.20.0)(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3))
       vite:
         specifier: 5.3.3
         version: 5.3.3(@types/node@20.14.10)(terser@5.31.3)
@@ -116,7 +116,7 @@ importers:
         version: 3.2.2(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3))
       vite-plugin-top-level-await:
         specifier: 1.4.1
-        version: 1.4.1(rollup@4.19.2)(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3))
+        version: 1.4.1(rollup@4.20.0)(vite@5.3.3(@types/node@20.14.10)(terser@5.31.3))
 
 packages:
 
@@ -292,8 +292,8 @@ packages:
     resolution: {integrity: sha512-mtsdpY1qyWgAO/iOK0L6gSGeR7GFcdW7tIjcNFxcWkfLDF5qVbPHKuGATFqRMsxcO8OUKNj0+3WOHB7EHm4Jdw==}
     engines: {node: '>=v18'}
 
-  '@commitlint/load@19.2.0':
-    resolution: {integrity: sha512-XvxxLJTKqZojCxaBQ7u92qQLFMMZc4+p9qrIq/9kJDy8DOrEa7P1yx7Tjdc2u2JxIalqT4KOGraVgCE7eCYJyQ==}
+  '@commitlint/load@19.4.0':
+    resolution: {integrity: sha512-I4lCWaEZYQJ1y+Y+gdvbGAx9pYPavqZAZ3/7/8BpWh+QjscAn8AjsUpLV2PycBsEx7gupq5gM4BViV9xwTIJuw==}
     engines: {node: '>=v18'}
 
   '@commitlint/resolve-extends@19.1.0':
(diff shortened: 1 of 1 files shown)


packages/alert/index.js
```diff
@@ -10,14 +8,16 @@ import '@warp-ds/icons/elements/warning-16';
...
+
 const variants = {
   negative: 'negative',
   positive: 'positive',
   warning: 'warning',
   info: 'info',
 };
 
-class WarpAlert extends WarpElement {
+class WarpAlert extends kebabCaseAttributes(WarpElement) {
```
BalbinaK, line 20: This shouldn't be necessary in components that have single-word properties/attributes. This wrapper ensures multiword kebab-cased attributes are mapped to camelCase properties. However, if you think it makes sense to  have this wrapper in case a new attribute is added in the future, then I'm fine with that 🙂

packages/alert/index.js
```diff
@@ -72,6 +49,23 @@ class WarpAlert extends WarpElement {
...
+  get _wrapperClasses() {
+    return classNames([ccAlert.wrapper, ccAlert[this.variant]]);
+  }
+
+  get _iconClasses() {
+    const activeIconClassNames = ccAlert[`${this.variant}Icon`];
+
+    return classNames([ccAlert.icon, activeIconClassNames]);
+  }
+
```
BalbinaK, line 68: I'm trying to understand this change. Is there a pattern you were thinking to introduce in our web components' structure? If so, what would that be?
My concern here is that the more changes are baked into one feature branch (`component-classes-cleanup`), the more difficult it will be for us to spot if we've made a mistake in that branch when merging with `next` branch.

packages/button/index.js
```diff
@@ -63,64 +63,77 @@ class WarpButton extends kebabCaseAttributes(WarpElement) {
...
     const primary = this.variant === 'primary';
+    return [
+      primary && !this.small && !this.quiet && !this.loading && ccButton.primary,
+      primary && this.small && !this.quiet && !this.loading && ccButton.primarySmall,
+      primary && this.small && this.quiet && !this.loading && ccButton.primarySmallQuiet,
+      primary && this.small && this.loading && (this.quiet ? ccButton.primarySmallQuietLoading : ccButton.primarySmallLoading),
+      primary && !this.small && this.quiet && !this.loading && ccButton.primaryQuiet,
+      primary && !this.small && this.loading && (this.quiet ? ccButton.primaryQuietLoading : ccButton.primaryLoading),
+    ];
+  }
```
BalbinaK, line 76: Just an idea I thought of now. Maybe we could move the `primary` condition to where `this._primaryClasses` is referenced, and so make the above conditions shorter + not repeat ourselves within the context of `_primaryClasses` getter? So something like this:
```
get _primaryClasses() {
  return [
    !this.small && !this.quiet && !this.loading && ccButton.primary,
    this.small && !this.quiet && !this.loading && ccButton.primarySmall,
    ...
  ]}
  
  
get _classes() {
  return classNames(this.buttonClass, [
      this.variant === 'primary' && ...this._primaryClasses,
      this.variant === 'secondary' && ...this._secondaryClasses,
      ...
  ])}
  ```

File: packages/toast/toast-container.js [hunk 4]
```diff
@@ -125,7 +127,7 @@ export class WarpToastContainer extends WarpElement {
   render() {
     return html`
       <aside class="${ccToaster.container}">
-        <div class="${ccToaster.toaster}" id="w-toast-container-list">
+        <div class="${ccToaster.base}" id="w-toast-container-list">
```
BalbinaK, line 130: I guess to make it consistent with the component name, this should be called `ccToastContainer` (and `toastContainer` in the css repo) but it's not a big deal :)

BalbinaK (commented):
Good job! Just a few comments 👇

packages/alert/index.js
felicia-haggqvist: ah that makes sense! 😄  I just added to it because I thought we wanted to have that for all of our components, but I can remove it! 🙂

packages/alert/index.js
felicia-haggqvist: Are you referring me moving this part below the static styles? I just moved them after static styles because I see that that is the structure we usually have in our web components, but I can revert that move. :)

packages/toast/toast-container.js
felicia-haggqvist: ah good catch! I'll change it :)

packages/alert/index.js
BalbinaK: It does make sense to keep a certain order of stuff in our web components. Perhaps we could also keep static styles and static properties closer to each other, too, and make the constructor come after those and before other getters/setters/methods/properties?
I'm just wondering if it makes sense to have these sort of changes in separate PRs (as a general rule) to facilitate the review process of big-a** PRs like `component-classes-cleanup -> next`.

packages/button/index.js
felicia-haggqvist: Ah nice! I'll change it to that :)

felicia-haggqvist: I have to change it like this instead 
`get _classes() {
    return classNames(this.buttonClass, [
      ...(this.variant === 'primary' ? this._primaryClasses : []),
      ...(this.variant === 'secondary' ? this._secondaryClasses : []),
      ...(this.variant === 'utility' ? this._utilityClasses : []),
      ...(this.variant === 'negative' ? this._negativeClasses : []),
      ...(this.variant === 'pill' ? this._pillClasses : []),
      ...(this.variant === 'link' ? this._linkClasses : []),
      this.fullWidth ? ccButton.fullWidth : ccButton.contentWidth,
    ]);
  }`

Otherwise it throws an error: "Expression expected"

packages/alert/index.js
felicia-haggqvist: Updated in this commit: [b7ce216](https://github.com/warp-ds/elements/pull/174/commits/b7ce216ee664d4d87d0c035cd0932d2f840c6255)

packages/toast/toast-container.js
felicia-haggqvist: Updated in this commit: [bbcc86d](https://github.com/warp-ds/elements/pull/174/commits/bbcc86da0f56532e0971b2329382b191740a1cd6)

packages/button/index.js
felicia-haggqvist: Updated in this commit: [26ae621](https://github.com/warp-ds/elements/pull/174/commits/26ae621926d669d1ebb9edc98a2cae74b10c09c6)

BalbinaK: We can actually do `this.variant === 'primary' && this._primaryClasses`. When I logged `this._classes` in the console, it showed a correct string. Seems like `classNames` function handles nested arrays.

felicia-haggqvist: Updated in this commit: [d6e1bf7](https://github.com/warp-ds/elements/pull/174/commits/d6e1bf724f3b23ad0801112f545a9327332ab033)

packages/affix/index.js
```diff
@@ -1,33 +1,30 @@
...
 
-import { fclasses } from '../utils';
-
-class WarpAffix extends WarpElement {
-  static styles = [WarpElement.styles];
+import { kebabCaseAttributes } from '../utils';
 
+class WarpAffix extends kebabCaseAttributes(WarpElement) {
   static properties = {
     ariaLabel: { type: String, attribute: 'aria-label' },
```
BalbinaK, line 14: If we use the `kebabCaseAttributes()` wrapper, the `attribute: 'aria-label'` option is redundant.

packages/textfield/index.js
```diff
@@ -1,12 +1,13 @@
...
+import { classNames } from '@chbphone55/classnames';
+import { input as ccInput, label as ccLabel, helpText as ccHelpText } from '@warp-ds/css/component-classes';
 import WarpElement from '@warp-ds/elements-core';
 import { ifDefined } from 'lit/directives/if-defined.js';
 
-import { classes } from '../utils';
+import { kebabCaseAttributes } from '../utils';
 
-class WarpTextField extends WarpElement {
+class WarpTextField extends kebabCaseAttributes(WarpElement) {
```
BalbinaK, line 10: Same here with the defined `attribute: 'some-attribute-name'` options in `static properties`. We can remove those in favour of the wrapper doing the same thing.

packages/toast/toast-container.js
```diff
@@ -14,7 +16,7 @@ import { repeat } from 'lit/directives/repeat.js';
  * @property   {Boolean}                              [canClose]  Whether the toast can be dismissed. Defaults to false. WARNING! For accessibility reasons, toasts should not be interactive and canclose should always be false. If the toast absolutely must be dismissble, set this to true.
  */
 
-export class WarpToastContainer extends WarpElement {
+export class WarpToastContainer extends kebabCaseAttributes(WarpElement) {
```
BalbinaK, line 19: `kebabCaseAttributes` wrapper is not needed here either.

packages/utils/expand-transition.js
```diff
@@ -4,7 +4,9 @@ import WarpElement from '@warp-ds/elements-core';
 import { collapse, expand } from 'element-collapse';
 import { ifDefined } from 'lit/directives/if-defined.js';
 
-class ExpandTransition extends WarpElement {
+import { kebabCaseAttributes } from './index';
+
+class ExpandTransition extends kebabCaseAttributes(WarpElement) {
```
BalbinaK, line 9: We don't need it here either because only `show` property is connected to a `show` attribute. The private property `_removeElement` doesn't need to be mapped to an attribute.

packages/toast/toast-container.js
BalbinaK: Same with SidebarElement and UnstyledHeading components.

packages/button/index.js
```diff
@@ -63,64 +63,70 @@ class WarpButton extends kebabCaseAttributes(WarpElement) {
...
+
+  get _pillClasses() {
+    return [
+      !this.loading && (this.small ? ccButton.pillSmall : ccButton.pill),
+      this.loading && (this.small ? ccButton.pillSmallLoading : ccButton.pillLoading),
+    ];
+  }
+
+  get _linkClasses() {
+    return [this.small ? ccButton.linkSmall : ccButton.link, !!this.href && ccButton.linkAsButton];
```
BalbinaK, line 118: I found a problem with this solution because we never enter this line when `link` is `false` while `href` has value:
Screenshot of a broken link that looks like a secondary button

We need to move `!!this.href && ccButton.linkAsButton` to line 128, and we can have `this.href && ccButton.linkAsButton` there.
Screenshot of a correct looking link that looks like a secondary button

felicia-haggqvist: Ah nicely spotted! 🙏  I'll move that part! 🎉

felicia-haggqvist: Updated in this commit: [04a29c8a](https://github.com/warp-ds/elements/pull/174/commits/04a29c8aadaf9281cd1cdc8a627548ed6cc3269a)

packages/affix/index.js
felicia-haggqvist: Updated in this commit: [5459cf7](https://github.com/warp-ds/elements/pull/174/commits/5459cf7eeceee45bb633efda91311217978ae8df)

packages/textfield/index.js
felicia-haggqvist: Updated in this commit: [5459cf7](https://github.com/warp-ds/elements/pull/174/commits/5459cf7eeceee45bb633efda91311217978ae8df)

packages/toast/toast-container.js
felicia-haggqvist: Updated in this commit: [9ddc7df](https://github.com/warp-ds/elements/pull/174/commits/9ddc7df079fcb7f4c661d5b5b6c0f8150dac7e89)

packages/utils/expand-transition.js
felicia-haggqvist: Updated in this commit: [9ddc7df](https://github.com/warp-ds/elements/pull/174/commits/9ddc7df079fcb7f4c661d5b5b6c0f8150dac7e89)

BalbinaK (approved):

PR #176, opened 2024-08-09
fix(select): autofocus to work in safari and firefox

felicia-haggqvist:
## Description
Noticed that the `autofocus` never seemed to have worked in Safari and FireFox for the Select component. This PR ensures that `autofocus` also works in Safari and Firefox, and fixes JIRA issue: [WARP-573](https://nmp-jira.atlassian.net/browse/WARP-573)

**Changes:**
- Removed `?autofocus=${this.autoFocus}`
- Added querySelector to focus Select element in `firstUpdated` for when `this.autoFocus` is `true` but `this.disabled` and `this.readOnly` are `false`.

**Tested:**
- Tested in Chrome, Safari and Firefox


packages/select/index.js
```diff
@@ -55,6 +55,11 @@ export class WarpSelect extends kebabCaseAttributes(WarpElement) {
     this._options = this.innerHTML;
   }
 
+  firstUpdated() {
+    // autofocus doesn't seem to behave properly in Safari and FireFox, therefore we set the focus here:
+    this.autoFocus && !this.disabled && !this.readOnly && this.shadowRoot.querySelector('select').focus();
```
BalbinaK, line 60: Does the component need to check if the other properties are falsy or should it be user's responsibility?

BalbinaK (commented):
Good catch! Should we perhaps point this PR directly to `next` instead of `component-classes-cleanup` branch? This way we'll separate out this fix in the release notes when `next` is merged to `main` and a new stable version of the package is released.

felicia-haggqvist:
> Good catch! Should we perhaps point this PR directly to `next` instead of `component-classes-cleanup` branch? This way we'll separate out this fix in the release notes when `next` is merged to `main` and a new stable version of the package is released.

@BalbinaK  Yeah I wasn't sure if we wanted this to be a part of `component-classes-cleanup`, but I agree that we can move this to be directly merged to next :)

packages/select/index.js
felicia-haggqvist: Hmm good question! 🤔 Yeah maybe it's enough if we just check for autoFocus here

BalbinaK:
> > Good catch! Should we perhaps point this PR directly to `next` instead of `component-classes-cleanup` branch? This way we'll separate out this fix in the release notes when `next` is merged to `main` and a new stable version of the package is released.
> 
> @BalbinaK Yeah I wasn't sure if we wanted this to be a part of `component-classes-cleanup`, but I agree that we can move this to be directly merged to next :)

You might need to cherry-pick this commit or wait until `component-classes-cleanup` is merged to `next` to minimise the diff (now it's 54 files changed).

felicia-haggqvist:
> > > Good catch! Should we perhaps point this PR directly to `next` instead of `component-classes-cleanup` branch? This way we'll separate out this fix in the release notes when `next` is merged to `main` and a new stable version of the package is released.
> > 
> > 
> > @BalbinaK Yeah I wasn't sure if we wanted this to be a part of `component-classes-cleanup`, but I agree that we can move this to be directly merged to next :)
> 
> You might need to cherry-pick this commit or wait until `component-classes-cleanup` is merged to `next` to minimise the diff (now it's 54 files changed).

@BalbinaK  Yes I was planning on waiting with merging this until `component-classes-cleanup` is merged to `next` :) I left a note in the PR description, but I can also change this PR into a draft-PR so it's clear to everyone that it is not ready to be merged. :)

packages/select/index.js
felicia-haggqvist: Removed `!this.disabled && !this.readOnly` in this commit: [de8c1c1](https://github.com/warp-ds/elements/pull/176/commits/de8c1c1db40fd7119aa48699fe6401ad51589de0)

BalbinaK (approved):

PR #182, opened 2024-08-15
Release 2.0.1

BalbinaK:


felicia-haggqvist (approved):

felicia-haggqvist:
Do we want to wait with making a release until we have a new stable release of @warp-ds/icons? Since we would need to also bump the @warp-ds/icons dependency in this repo? @imprashast

PR #183, opened 2024-08-15
fix(badge): remove deprecated 'notification' example

BalbinaK:
The notification variant of Badge is no longer supported in Warp and a corresponding token was removed from @warp-ds/css v2.0.0. This PR removes "notification" variant from Badge properties and cleans up the docs examples.


felicia-haggqvist (approved):
🎉

PR #190, opened 2024-11-15
fix: Add select element change event

adama-schibsted:
https://nmp-jira.atlassian.net/browse/WARP-647.

Adds a custom change event to the select component, which fires on dropdown change.

The dropdown value is available on the `detail` field.

To use in Svelte and other libraries (without a wrapper), add event listeners manually (using addEventListener).

### DIFF
--- a/packages/select/index.js
+++ b/packages/select/index.js
@@ -93,8 +93,8 @@ export class WarpSelect extends kebabCaseAttributes(WarpElement) {
   }
 
   // Fire a custom 'change' event, used when the dropdown changes state.
-  onChange() {
-    let event = new CustomEvent('change', { detail: 'change' });
+  onChange(value) {
+    let event = new CustomEvent('change', { detail: value });
 
     this.dispatchEvent(event);
   }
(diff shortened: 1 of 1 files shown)


packages/select/index.js
```diff
@@ -120,7 +127,8 @@ export class WarpSelect extends kebabCaseAttributes(WarpElement) {
           aria-describedby="${ifDefined(this.#helpId)}"
           aria-invalid="${ifDefined(this.invalid)}"
           aria-errormessage="${ifDefined(this.invalid && this.#helpId)}"
-          @keydown=${this.handleKeyDown}>
+          @keydown=${this.handleKeyDown}
+          @change=${(e) => this.onChange(e.target.value)}>
```
imprashast, line 131: Maybe cleaner if you just provide the function here as a callback `@change=${this.onChange}` and then set the value in the function as `{ detail: e.target.value }`

adama-schibsted: Nice! Updated it.

PR #193, opened 2024-11-22
fix: Improve attention component styling

adama-schibsted:
Improve attention component by running the recompute function on render.
Ensures that arrow/highlight position is computed correctly on first render, giving correct styling if the highlight is shown by default.

### DIFF
--- a/packages/attention/index.js
+++ b/packages/attention/index.js
@@ -98,7 +98,11 @@ class WarpAttention extends kebabCaseAttributes(WarpElement) {
     }
 
     // Fix FOUC effect issues
-    setTimeout(() => this.requestUpdate(), 0);
+    setTimeout(() => {
+      this.requestUpdate();
+      this.handleDone(); // Run handleDone initially, to compute correct arrow position etc. directly.
+    }, 0);
+
     if (!this.callout) {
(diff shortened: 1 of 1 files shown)


felicia-haggqvist:
> Not sure it's related, but I see that there's another `recompute` call resulting from `this.requestUpdate()`. It gets triggered at the end of `updated()` method ([see code](https://github.com/warp-ds/elements/blob/next/packages/attention/index.js#L241-L243)), which is the last method in the [update lifecycle](https://lit.dev/docs/components/lifecycle/#reactive-update-cycle). `handleDone` triggers a `recompute`, too, so I was wondering if any of the two would be redundant here.

I think we need to keep both `recompute` calls. 🤔  If we remove this `recompute` from the `updated()` method, then the repositioning of both the attention-element and the arrow-element could jump and get misaligned. See video for example: 

https://github.com/user-attachments/assets/568b85c1-ad70-4047-baff-ba631d7cfb20

felicia-haggqvist (approved):

BalbinaK (approved):

adama-schibsted:
Good points! I'll merge.

PR #201, opened 2024-12-04
fix: ease strictness of warp-ds/css peer dependency

wkillerud:
Allow any 2.x

Fixes a peer dependency warning when I have `2.0.1`

```
 WARN  Issues with peer dependencies found
.
└─┬ @warp-ds/elements 2.0.1
  └── ✕ unmet peer @warp-ds/css@2.0.0: found 2.0.1
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   },
   "license": "Apache-2.0",
   "peerDependencies": {
-    "@warp-ds/css": "2.0.0",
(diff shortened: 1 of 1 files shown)


imprashast:
I totally forgot to check this but this should not be merged.

The whole point of this is that a certain version of elements package depends on a certain version of css, hence this is so strict. Making it 2.x defies the whole point of this.

PR #216, opened 2025-08-20
fix: add form participation to w-button

digitalsadhu:
This adds form participation to w-button so that if a name and value are provided, these are captured by the form. Additionally, if the button is type="submit" it will submit the parent form and if the type is "reset" it will reset the parent form


packages/button/index.js
```diff
@@ -16,7 +17,7 @@ import { messages as svMessages } from './locales/sv/messages.mjs';
 
 const buttonTypes = ['primary', 'secondary', 'negative', 'utility', 'pill', 'link'];
 
-class WarpButton extends kebabCaseAttributes(WarpElement) {
+class WarpButton extends FormControlMixin(kebabCaseAttributes(WarpElement)) {
```
digitalsadhu, line 20: Add the open wc mixin so we dont need to implement our own. This provides some hooks and a `this.setValue` method for us and sets up the element internals under the hood

packages/button/index.js
```diff
@@ -34,10 +35,18 @@ class WarpButton extends kebabCaseAttributes(WarpElement) {
     rel: { type: String, reflect: true },
     fullWidth: { type: Boolean, reflect: true },
     buttonClass: { type: String, reflect: true },
+    name: { type: String, reflect: true },
+    value: { type: String, reflect: true },
```
digitalsadhu, line 39: Every element that is to be associated with a form needs to support name and value

packages/button/index.js
```diff
@@ -34,10 +35,18 @@ class WarpButton extends kebabCaseAttributes(WarpElement) {
...
   };
 
   static styles = [WarpElement.styles];
 
+  updated(changedProperties) {
+    if (changedProperties.has('value')) {
+      this.setValue(this.value);
+    }
+  }
```
digitalsadhu, line 48: any time value is set or changed, either via the attribute or the property, we need to update the form internals using the setValue method provided by the open wc mixin. so Long as we do this, the form will capture name and value for us if set.

PR #222, opened 2025-09-01
test: add vitest, playwright, example test

wkillerud:
Add a test setup with vitest


vitest.config.ts
```diff
@@ -0,0 +1,34 @@
+import { defineConfig } from "vite";
```
pearofducks, line 1:
```suggestion
import { defineConfig } from "vitest/config";
```

PR #239, opened 2025-09-01
chore: try to fix preview

wkillerud:
Asset loading is broken in previews because of wrong base during build


wkillerud:
Hard-coded paths in so many places, pausing this for now. Maybe we want to rethink the whole dev server/preview thing.

PR #273, opened 2025-09-03
chore: migrate select component

wkillerud:
Fix a type issue in Buttons while at it

### DIFF
--- a/packages/select/index.ts
+++ b/packages/select/index.ts
@@ -11,13 +11,13 @@ import { when } from 'lit/directives/when.js';
 
 import { activateI18n } from '../i18n.js';
 import { reset } from '../styles.js';
-import { styles } from './styles.js';
 
 import { messages as daMessages } from './locales/da/messages.mjs';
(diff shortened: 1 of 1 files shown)


packages/select/index.ts
```diff
@@ -20,46 +20,57 @@ import { messages as svMessages } from './locales/sv/messages.mjs';
...
-    invalid: { type: Boolean, reflect: true },
+  // Renders the field in an invalid state. Often paired with `hint` to provide feedback about the error
+  @property({ type: Boolean, reflect: true })
+  invalid: boolean;
 
-    // Whether to always show a hint
-    always: { type: Boolean, reflect: true },
+  // Whether to always show a hint
+  @property({ type: Boolean, reflect: true })
+  always: boolean;
```
joms, line 34: I think the name is a bit vague

```suggestion
  showHint: boolean;
```

wkillerud: No breaking changes this round

PR #275, opened 2025-09-03
chore: updating pill component

### LINKED ISSUE #270: Pill

leftieFriele:
Close #270


wkillerud:
I think some of the names could use an alternative. Maybe open an issue with that discussion? We won't do any breaking changes for now so we can get this out the door.

leftieFriele:
> I think some of the names could use an alternative. Maybe open an issue with that discussion? We won't do any breaking changes for now so we can get this out the door.

good idea, I started on one here https://github.com/warp-ds/elements/issues/297

PR #278, opened 2025-09-03
chore: upgrade attention

flatoy:
convert to ts
remove kebabcase
use LitElement
inline styles


wkillerud:
Would love it if you could add Storybook stories for this too.

Pretty much add these: https://github.com/warp-ds/elements/blob/next/pages/components/attention.html 

There's one example you can look at (we need stuff from the script tag at the bottom for it to work)

PR #299, opened 2025-09-03
docs: use the custom element manifest for docs

wkillerud:
Add autodocs as well for the nice previews of all variants


wkillerud:
I want to go through and add a `Component.mdx` where relevant and test that all the examples behave as I expect, but this is in a good enough shape to ship for now.

PR #304, opened 2025-09-05
chore(storybook): fix some story behavior

pearofducks:
- The 'spread' stuff didn't seem to be working on dead-toggle's stories, so just wired it directly
- Modal had some missing images and bad styling using Unocss classes, just used style directly instead


wkillerud (approved):
Sweet, thanks!

wkillerud:
Ah, `spread` [expects the attribute key to include the Lit syntax](https://open-wc.org/docs/development/lit-helpers/#spread), it doesn't do any magic for us there.

PR #305, opened 2025-09-05
chore: fix Lit syntax before using `spread`

wkillerud:
Fixes the issue we saw of boolean controls not turning off from #304 for example


pearofducks (approved):
Fiiiine fix the root issue instead of having crappy hacks laying around, I guess that's "better" 😆

PR #306, opened 2025-09-05
docs: generate editor helpers, add instructions for use

wkillerud:
If you use VS Code I set up the configuration here in the workspace. Try it out in any HTML file!


index.html
wkillerud, line 1: Yes, I opened this file, how can you tell? 🙈

PR #314, opened 2025-09-10
feat: add WarpPageIndicator component with styles and stories

imprashast:


packages/pageindicator/pageindicator.stories.tsx
```diff
@@ -0,0 +1,78 @@
+import { html, LitElement } from 'lit';
+import './index.ts';
+
+export default {
+  component: 'w-pageindicator',
+  title: 'PageIndicator',
```
wkillerud, line 6: Another nit, feels like this could be grouped with Pagination and such in the sidebar
```suggestion
  title: 'Navigation/PageIndicator',
```

PR #316, opened 2025-09-10
fix: adding link component

leftieFriele:
This is the first step towards splitting Button into Button and Link. For now, we use the Link inside the Button component.

- removes the loading state and loading text
- adds two quiet variants
Closes https://github.com/warp-ds/elements/issues/284


packages/button/Button.mdx
```diff
@@ -48,7 +48,7 @@ Buttons used in the context of a form.
 
 ## Link (anchor) styled as button
 
-Buttons will be rendered as an anchor (a tag) if they use an href attribute.
+See [Link](/docs/buttons-link--docs) component for details.
```
leftieFriele, line 51: Perhaps we should suggest using `w-link` as well, not just point to it? 🤔

PR #317, opened 2025-09-10
chore: migrate datepicker from React

wkillerud:
Still need to add tests, docs and figure out some TODOs around localization. But it seems to be working (he said with a question mark in his voice).

https://github.com/user-attachments/assets/292b1e0d-5053-4bd6-84a8-39fbb5588d9b

### DIFF
--- a/.cem/custom-element-manifest.config.js
+++ b/.cem/custom-element-manifest.config.js
@@ -8,7 +8,13 @@ export default {
   },
 
   // Globs to analyze
-  globs: ['packages/**/index.ts', 'packages/**/index.js', 'packages/modal/modal-*.ts', 'packages/toast/toast*.js'],
+  globs: [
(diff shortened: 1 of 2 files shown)


packages/datepicker/index.ts
```diff
@@ -0,0 +1,428 @@
...
+            @click="${this.#toggleCalendarOpen}">
+            <w-icon-calendar-16></w-icon-calendar-16>
+          </w-button>
+        </div>
+      </div>
+      ${this.isCalendarOpen
+        ? html`
+            <div class="w-dropdown__popover w-dropdown__popover--open">
+              <div
+                aria-roledescription="TODO"
```
wkillerud, line 342: There are a couple of these text snippets that are part of a `phrases` object in the React version. How do we want to deal with these?

PR #329, opened 2025-09-22
feat: add slider and range slider implementations

wkillerud:
- Native range input field for formy goodness.
- CSS anchor positioning to give our tooltip component a target DOM element that follows the range input thumb, without all the JS maths.
- OK, there's still some math in CSS to calculate the fill width using custom properties, but it's all to calculate a percentage of the range to fill.
- Input masking when the field is blurred so as to play nice with screen readers.

Tasks:

- [x] Fix range slider style.
- [x] Test a bunch of different slider values and see if the CSS math holds up.
- [x] Cross-browser visual QA.
- [x] (Developer) Accessibility testing.
- [x] Opt-in visual markers below slider track.
- [x] Validation error on bad input in textfield.
- [x] Prevent from value and to value from crossing each other.
- [x] Prevent from and to values from overlapping visually.
- [x] Write unit tests.

### DIFF
--- a/packages/textfield/textfield.test.ts
+++ b/packages/textfield/textfield.test.ts
@@ -1,13 +1,92 @@
 import { html } from 'lit';
 
-import { expect, test } from 'vitest';
+import { userEvent } from '@vitest/browser/context';
+import { expect, test, vi } from 'vitest';
 import { render } from 'vitest-browser-lit';
 
+import '../affix/index.js';
 import './index.js';
 
 test('renders the textfield', async () => {
-  const component = html`<w-textfield label="Email" help-text="Ugyldig e-post"></w-textfield>`;
+  const component = html`<w-textfield label="Test label"></w-textfield>`;
 
   const page = render(component);
-  await expect.element(page.getByText('Email')).toBeVisible();
+  await expect.element(page.getByText('Test label')).toBeVisible();
+});
+
+test('works as expected in forms', async () => {
+  const label = 'Test label';
+  const component = html`
+    <form data-testid="form">
+      <w-textfield label="${label}" name="message" value="Hola el Mundo"></w-textfield>
+    </form>
+  `;
+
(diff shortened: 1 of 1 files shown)


wkillerud:
The `preventDefault()` trick to stop the slider thumbs from crossing adds this weird quirk where if you're quick, you get stopped before the overlap happens. Not quite what I had in mind.

https://github.com/user-attachments/assets/19598d70-3fdf-4065-b8f3-deb84445458c

wkillerud:
We'll include the anchor positioning polyfill in browsers that need it. Looking into how we can do that without shipping it to browsers that have native support.

https://www.npmjs.com/package/@oddbird/css-anchor-positioning

packages/slider/slider-thumb.ts
```diff
@@ -0,0 +1,323 @@
...
+    return true;
+  }
+
+  async connectedCallback() {
+    super.connectedCallback();
+    if (!('anchorName' in document.documentElement.style)) {
+      // Load the polyfill for CSS anchor positioning by @oddbird for browsers without native support.
+      const dirname = import.meta.url.substring(0, import.meta.url.lastIndexOf('/'));
+
+      const [{ default: polyfill }] = await Promise.all([
```
imprashast, line 172: Should you handle promise rejection here? Else it'll be silent and impossible to debug, if it ever fails :)

packages/textfield/index.ts
```diff
@@ -194,33 +213,44 @@ class WarpTextField extends FormControlMixin(LitElement) {
   render() {
     return html`
       ${this._label}
-      <div class="${ccinput.wrapper}">
+      <div
```
imprashast, line 216: I'm assuming you have checked that none of the textfield changes break current API's?

imprashast (approved):
Nice, Good job 👍

PR #330, opened 2025-09-25
Add tabs

imprashast:


packages/tabs/tabs.ts
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  render() {
+    const navClasses = classNames(this.tabClass, ccTabs.wrapper);
+    const divClasses = classNames([ccTabs.base, this._gridClass]);
+
+    return html`
+      <div class="${navClasses}">
+        <div role="tablist" class="${divClasses}" @keydown="${this._handleKeyDown}">
+          <slot></slot>
```
digitalsadhu, line 260: would it make sense to also run update tabs when the slot changes in case w-tabs are added or removed?

digitalsadhu (approved):

PR #344, opened 2025-10-02
Fix dev setup

digitalsadhu:
This should fix up intellisense in React apps.


packages/affix/index.ts
```diff
@@ -120,3 +120,10 @@ declare global {
     'w-affix': WarpAffix;
   }
 }
+declare module 'react' {
```
imprashast, line 123: Shouldn't this belong in the `react.ts` files as they are only to be used by people using react and not the web-component itself?

package.json
```diff
@@ -74,9 +75,8 @@
   "peerDependencies": {
     "@lingui/core": "5.x",
     "@warp-ds/css": "2.1.1",
-    "@warp-ds/elements-core": "2.x",
     "lit": "3.x",
-    "react": "^19.1.1"
+    "react": "^18.x"
```
imprashast, line 79: Maybe react should not be a peerDependency here as the team who is only using this repo with Lit, would get react in their node modules because of this :)

wkillerud (approved):
Sweet! Love me some code suggestions.

packages/affix/index.ts
digitalsadhu: I've moved all of this to a root types file now to get intellisense to play nice.
We need it to support both React with wrappers as well as React 19 without wrappers.

package.json
digitalsadhu: yeah makes sense!

digitalsadhu: I'll remove all peer deps. We just want the end users to just install the packages and get intellisense so no need to make them install Lit or Warp CSS etc either.

PR #359, opened 2025-10-22
fix: decouple react wrappers

digitalsadhu:
Adds a storybook setup for React wrappers and stories for each component's react wrapper. 
Various fixes along the way as discovered to ensure React wrappers work correctly.
Also decouples the React wrappers with the implementation so that the same production bundle can be shared for both CE or React wrappers

### DIFF
--- a/packages/affix/react.ts
+++ b/packages/affix/react.ts
@@ -1,14 +1,18 @@
-import { LitElement } from 'lit';
 import { createComponent } from '@lit/react';
+import { LitElement } from 'lit';
 import React from 'react';
 
 import { WarpAffix } from './index';
 
-// decouple from CDN by providing a dummy class
+// We do this in order to prevent users from bundling in the CE implementation when
+// they should be getting it from Eik.
 class Component extends LitElement {}
 
 export const Affix = createComponent({
   tagName: 'w-affix',
+  // This tricks TS into thinking the stub we created above is still the actual implementation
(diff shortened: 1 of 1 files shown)


pearofducks:
Would it make sense somewhere to explain these two patterns?

```ts
// decouple from CDN by providing a dummy class
class Component extends LitElement {}
```

```ts
Component as unknown as typeof WarpCombobox,
```

I have some idea what the former does and why we have it - but might be good to explicitly document. The latter I have absolutely no idea what happens when you chain `as` like that - maybe I'm just a TS noob?

digitalsadhu:
> Would it make sense somewhere to explain these two patterns?
> 
> ```ts
> // decouple from CDN by providing a dummy class
> class Component extends LitElement {}
> ```
> 
>  ```ts
> Component as unknown as typeof WarpCombobox,
> ```
> 
> I have some idea what the former does and why we have it - but might be good to explicitly document. The latter I have absolutely no idea what happens when you chain `as` like that - maybe I'm just a TS noob?

Yup, probably a good idea to explain it.

The former breaks the inlining of the code into the wrappers to prevent users from bundling in the CE version when they should be getting it from Eik. The latter just tricks TS into thinking the dumb stub is still the actual class we want Devs IDEs to think it is. This way devs can import the react wrapper and use it while still getting the full intellisense experience yet not bundling in the CE implementation into their app bundles. I can add something to one of the components.

PR #408, opened 2025-11-05
fix: restrict components to release to just existing

digitalsadhu:
Restrict to only ship the existing components


Skadefryd (approved):
Nice, Should there be 15 in this list?

PR #419, opened 2025-11-12
fix: add components back in post release

digitalsadhu:


package.json
```diff
@@ -19,10 +19,8 @@
     "./components/card": "./dist/packages/card/index.js",
     "./components/expandable": "./dist/packages/expandable/index.js",
     "./components/modal": "./dist/packages/modal/index.js",
-    "./components/pagination": "./dist/packages/pagination/index.js",
```
digitalsadhu: removed these as the /components path is just there for backwards compatibility for the existing 14 components

PR #420, opened 2025-11-12
fix: use explicit exports to support module resolution from 🦖-age

joms:


digitalsadhu (approved):
Im with William, if it works it works

PR #438, opened 2025-11-20
fix: checkbox styles

### LINKED ISSUE #429: Checkbox
https://nmp-jira.atlassian.net/browse/WARP-1236

Findings:
Critical

* [x] Not selected needs to use Border/Strong token - currently uses Border.
 
Bugs:
* [x] Can users change horizontal gap between checkboxes?
    * Discussion in warp-bug-hunt slack channel. Vertical gap was intended here, between the checkboxes/radios, and it should be 16px
* [x] Cannot check Error and Disabled states, unavailable in Storybook.

https://miro.com/app/board/uXjVLmvdEfA=/?moveToWidget=3458764644035266031&cot=14

pearofducks:
- Adds a `w-checkbox-group` element to provide 16px spacing between elements (design has also hinted this might need to be user-controllable in the future, so it uses `part`).
- Fixes other UI bugs to close #429

### DIFF
--- a/.cem/custom-element-manifest.config.js
+++ b/.cem/custom-element-manifest.config.js
@@ -39,6 +39,7 @@ export default {
     'packages/pagination/index.ts',
     'packages/rip-and-tear-radio/radio*.ts',
     'packages/rip-and-tear-checkbox/checkbox.ts',
+    'packages/rip-and-tear-checkbox/checkbox-group.ts',
     'packages/slider/slider.ts',
     'packages/slider/slider-thumb.ts',
     'packages/steps/index.ts',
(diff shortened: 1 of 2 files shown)


digitalsadhu (changes requested):
We need a react wrapper for this
We also need to add it to the cem list of components to be scanned and tsconfig as well as the package json exports

* https://github.com/warp-ds/elements/blob/b490c448a2a4e122adaa9abc04ecc3d78640bb4d/.cem/custom-element-manifest.config.js#L41
* https://github.com/warp-ds/elements/blob/b490c448a2a4e122adaa9abc04ecc3d78640bb4d/package.json#L96
* https://github.com/warp-ds/elements/blob/b490c448a2a4e122adaa9abc04ecc3d78640bb4d/tsconfig.json#L26

And yes we can definitely improve this but it got more explicit due to trying to exclude some components for a release while including others.

wkillerud (approved):

PR #448, opened 2025-11-24
fix: updated styling

Skadefryd:
Figma
image

Web implementation
image
image


Skadefryd:
uses the new border-strong, and border-strong-hover for the borders, and also the handle itself in the off states.

PR #452, opened 2025-12-01
fix: rename step indicator

digitalsadhu:
* Renames steps to step indicator
* accessibility adjustment for parent/child list


packages/steps/react.ts
```diff
@@ -1,14 +1,14 @@
 import { LitElement } from 'lit';
 import { createComponent } from '@lit/react';
 import React from 'react';
-import { WarpStep, WarpSteps } from './index.js';
+import { WarpStep, WarpStepIndicator } from './index.js';
 
 // decouple from CDN by providing a dummy class
 class Component extends LitElement {}
 
 export const Steps = createComponent({
```
wkillerud: We should rename the React component as well for consistency
```suggestion
export const StepIndicator = createComponent({
```

wkillerud (approved):

PR #453, opened 2025-12-02
fix: implement error messages in sliders

joms:
The idea is that an error should always be a string. `""` is the same as `error = false`, and if it contains something it should be the same as true. I think it would make sense if slider-thumb `invalid` is just a boolean, while in `slider` it is represented as a string. However, I got lost in how the `invalid` state is passed back and forth...

image

Current issues:
- ~Error message from thumbs are reset on the way out. I suspect there's an event that is triggered and clears the error. Can be seen by adding `console.log(e.detail)` to `#onSliderValidity` in `slider.ts`~
- ~External error doesn't set thumbs to invalid.~

### DIFF
--- a/packages/slider/slider.ts
+++ b/packages/slider/slider.ts
@@ -258,7 +258,6 @@ class WarpSlider extends FormControlMixin(LitElement) {
         @input="${this.#onInput}"
         @slidervalidity="${this.#onSliderValidity}"
         aria-invalid="${this.errorText ? 'true' : nothing}"
-        aria-describedby="${this.errorText ? 'error-slot' : nothing}"
       >
         ${
           this.label


packages/slider/slider-thumb.ts
```diff
@@ -408,14 +440,14 @@ class WarpSliderThumb extends FormControlMixin(LitElement) {
...
-      );
-    }
+    // if (changedProperties.has('invalid')) {
+    //   this.dispatchEvent(
+    //     new CustomEvent('slidervalidity', {
+    //       bubbles: true,
+    //       detail: { invalid: this.invalid },
+    //     }),
+    //   );
+    // }
```
wkillerud:
```suggestion
```

trygve-lie:
I'm a bit from the outside here:

>  on this new API 

I see we are using form association in these components. Great. But are we extending the native form validation or are we rolling our own?

I'm looking in the direction of the [Constraint Validation API](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Constraint_validation) and [Validity State](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState).

PR #464, opened 2025-12-10
fix: align filenaming across components

digitalsadhu:
1. gets rid of "rip and tear"
2. removes - characters from names: eg. dead-toggle -> deadtoggle
3. renames steps to stepindicator


imprashast (approved):
RIP rip-and-tear

wkillerud (approved):

PR #472, opened 2025-12-23
fix: separate out Eik bundles by language

digitalsadhu:
**Full disclosure**: I tried out Claude Code to implement this. It was able to plan and implement the changes asked for. Whether the code is understandable or the approach is sound, I leave that to the astute reader to decide.

**What this PR does:**
Produce a bundle for each language as well as the original multi lingual bundle for backwards compatibility. Each language now has its own route. So the bundle files are now:
* /index.js - backwards compatibility, switches on html lang attr
* /en/index.js
* /nb/index.js
* /sv/index.js
* /da/index.js
* /fi/index.js

**How it does it:**
1. takes in language as an env var to the build, if not set, old behaviour is preserved.
2. adds a custom rollup plugin that stubs out all the languages that are not the desired language for the current build (so that things don't break)
3. modifies i18n file to use env vars and detectLocale to return short circuit and return locale from env var if provided.
4. uses the usual rollup method to inline env vars in the build
5. modify build commands to run a build for each language, setting the env var for each.

**Discussion:**
- 9kb saving in the per language bundles. It's not much but it's something. There is another 5kb potential savings if we were to switch to lit localise instead of lingui.
- the main decision point probably hangs off of whether switching language by url rather than html language tag is desired.

### DIFF
--- a/build/bundle-all-locales.js
+++ b/build/bundle-all-locales.js
@@ -1,0 +1,55 @@
+#!/usr/bin/env node
+
+/**
+ * Build script that produces:
+ * 1. Per-language bundles: eik/{locale}/index.js for each supported locale
+ * 2. Backwards-compatible all-languages bundle: eik/index.js
+ *
+ * This runs the Rollup build multiple times with different WARP_LOCALE values.
+ */
+
+import { execSync } from 'node:child_process';
+import { existsSync, mkdirSync } from 'node:fs';
+
+const supportedLocales = ['en', 'nb', 'fi', 'da', 'sv'];
+const outdir = new URL('../eik', import.meta.url).pathname;
+
+// Ensure output directories exist
+for (const locale of supportedLocales) {
+  const localeDir = `${outdir}/${locale}`;
+  if (!existsSync(localeDir)) {
+    mkdirSync(localeDir, { recursive: true });
+  }
+}
+
+console.log('Building per-language bundles...\n');
+
+// Build per-language bundles
+for (const locale of supportedLocales) {
+  console.log(`Building bundle for locale: ${locale}`);
+  try {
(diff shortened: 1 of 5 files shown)


digitalsadhu:
> My only worry is that we'll end up with a mix of references to `/index.js` and `/<locale>/index.js` on the same page, defeating the purpose.
> 
> I know we include Elements in the base HTML template as a global, so that's all good. It might not be a problem, but I worry folks still import Warp elements and get it import-mapped out to the generic `/index.js` by Eik.

yeah, theres definitely an element of its going to get worse before it gets better and a question in my mind of whether its worth it.
I'd quite like @trygve-lie to chime in on this specifically.

trygve-lie:
I think this is a good thing but perhaps the wrong time to introduce this.

On our higher order road map for next year we would like to introduce multi language support for end users and making it possible for end users to switch language on a marketplace.

By providing such a feature we need to look at how we do this and have a strict way of loading language in the apps. We will probably also need to have a team wide joined effort to get multi language support out there in the apps. In making a strict way for how this is done, we should be able to reduce the risk of duplicate loading.

I am kinda thinking that this might be good to post pone until we start with adding multi language support for end users.

digitalsadhu:
> I think this is a good thing but perhaps the wrong time to introduce this.
> 
> On our higher order road map for next year we would like to introduce multi language support for end users and making it possible for end users to switch language on a marketplace.
> 
> By providing such a feature we need to look at how we do this and have a strict way of loading language in the apps. We will probably also need to have a team wide joined effort to get multi language support out there in the apps. In making a strict way for how this is done, we should be able to reduce the risk of duplicate loading.
> 
> I am kinda thinking that this might be good to post pone until we start with adding multi language support for end users.

Sounds good to me. We can leave the branch around for reference.

PR #490, opened 2026-01-20
fix: use default slot in w-tab

digitalsadhu:
1. use default slot instead of label property in w-tab
2. fix bug with panels sliding up to the right of the tabs
3. add slotted style to help icons behave 
4. fix a bug where clicking the panels triggered a console error
5. remove ability to provide additional styles via properties (will be replaced with :part styling once we land on patterns for this)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [2.4.0-next.8](https://github.com/warp-ds/elements/compare/v2.4.0-next.7...v2.4.0-next.8) (2026-01-21)
+
+
+### Features
+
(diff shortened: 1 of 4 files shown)


wkillerud (approved):
terrific

packages/tab/tab.ts
```diff
@@ -23,33 +23,27 @@ const ccButtonReset = 'focus:outline-none appearance-none cursor-pointer bg-tran
  * [See Storybook for usage examples](https://warp-ds.github.io/elements/?path=/docs/tabs--docs)
  */
 export class WarpTab extends LitElement {
-  static styles = [reset, styles];
+  static styles = [reset, styles, css`::slotted([slot="icon"]){display:flex}`];
```
imprashast: Isn't it better to add this to the icon itself than here?

PR #492, opened 2026-01-21
feat: radio rework

digitalsadhu:
After a discussion thread with Trygve, I thought I would try reimplementing the radio (mostly it was Claude) without the usual hacks to get it styled correctly. The result is visually the same (I think). It's about 200 LOC (mostly due to various accessibility features, keyboard nav etc.) vs maybe a 1000 for the web awesome one though it's not totally the same either. Might just be an interesting exercise but I push it here dear reader for your perusal should you so feel inclined.

### DIFF
--- a/packages/radio-basic/radio-basic.ts
+++ b/packages/radio-basic/radio-basic.ts
@@ -296,8 +296,9 @@ export class WarpRadioBasic extends BaseFormAssociatedElement {
   render() {
     return html`
       <input
-        id=${this.#uniqueId}
         type="radio"
+        part="control"
+        id=${this.#uniqueId}
         .name=${this.name ?? ''}
         .value=${this.value}
         .checked=${this.checked}
@@ -305,7 +306,7 @@ export class WarpRadioBasic extends BaseFormAssociatedElement {
         ?required=${this.required}
         @change=${this._handleChange}
       />
-      <label for=${this.#uniqueId}><slot></slot></label>
+      <label part="label" for=${this.#uniqueId}><slot></slot></label>
     `;
   }
 }
--- a/packages/radio-basic/styles.ts
+++ b/packages/radio-basic/styles.ts
@@ -4,96 +4,117 @@ export const styles = css`
   :host {
     display: inline-flex;
     align-items: center;
-    gap: 8px;
+    gap: var(--_gap);
+
+    --_gap: var(--w-c-radio-gap, 8px);
+
+    --_size: var(--w-c-radio-size, 2rem);
+    --_radius: var(--w-c-radio-radius, 50%);
+    --_border-width: var(--w-c-radio-border-width, 1px);
+    --_checked-border-width: var(--w-c-radio-checked-border-width, 0.6rem);
+
+    --_bg: var(--w-c-radio-bg, var(--w-s-color-background));
+    --_border-color: var(--w-c-radio-border-color, var(--w-s-color-border-strong));
+    --_border-color-checked: var(--w-c-radio-border-color-checked, var(--w-s-color-border-selected));
+
+    --_outline-width: var(--w-c-radio-outline-width, 2px);
+    --_outline-color: var(--w-c-radio-outline-color, var(--w-s-color-border-focus));
+    --_outline-offset: var(--w-c-radio-outline-offset, var(--w-outline-offset, 1px));
+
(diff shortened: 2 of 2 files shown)


trygve-lie:
Awesome!

A good chunk of code here and a bit difficult to reason on the code vs the current existing implementation without doing a bit more detailed comparison.  But I do think the internal code of the component are secondary here.

This kinda come from a discussion (in a private slack channel) about layered dependencies in components and how we many times throw a lot of the native HTML elements out and re-implement existing elements from the ground up just so we can apply what ever design we want to the element. The discussion is then; how do we balance design requirements against how much code the feature require on these native HTML elements?

> though it's not totally the same either.

If we ignore the code here; what are the differences? Is this fulfilling the design requirements, the required functionality and accessibility?

Skadefryd:
Added a little commit redoing the styling setup, the concept is up for discussion and is explained in this long winded comment below:   (This is a iteration of the same setup already in use in some of these components ).  We could add support for overriding stuff like this to our UNOcss drive as well if we want to.  Also worth mentioning theres a few component specific tokens coming from Figma, these follow the same pattern as the semantic tokens when it comes to implementation.  

# Styling / override model 
 
> Oh so you are special and absolutely HAVE to customize a component?  this is how!.

This component uses a layered styling approach so we can support theming, variants and one-off tweaks without duplicating selectors or coupling consumers to internals.

There are a few different ways to influence styling, and they’re intentionally not all equal.

## 1. Semantic tokens (--w-s-*)

These come from the design system (Figma → Style Dictionary → CSS vars).  

### How they’re used:

- They are defaults only
- The component does not style itself directly from --w-s-*
- They’re mapped once in :host and then forgotten about

### Example (inside the component):

`--_border-color: var(--w-c-radio-border-color, var(--w-s-color-border-strong));`


### Pros

- Stable, brand-aware
- Theme changes “just work”
- Good long-term foundation

### Cons

- Not granular enough for component-specific 'special' needs

### Used for
Brand meaning (text, borders, backgrounds, states)


## 2. Component tokens (--w-c-radio-*)

This is the public styling API of the component.

### How they’re used:

- Mapped to internal tokens (--_*) in :host
- Can be set by variants, states, or from outside the component
- All selectors only read from internal tokens

### Example (consumer usage):

```
.settings-panel w-radio {
  --w-c-radio-size: 1.5rem;
  --w-c-radio-border-color-checked: var(--w-s-color-border-selected);
}

```

Example (variant/state inside the component):

```
:host([disabled]) {
  --w-c-radio-border-color: var(--w-c-radio-border-color-disabled);
  --w-c-radio-label-color: var(--w-c-radio-label-color-disabled);
}
```


### Pros

- Explicit and discoverable
- Works for contextual theming (based on where the component is used)
- No selector duplication
- Safe refactors later (selectors can change, tokens stay)

### Cons

- Every exposed token is API surface
- Needs discipline to keep the set small

### Use for

- Size, spacing, colors, emphasis
- Variants and states
- Product-level theming


## 3. Parts (::part(control), ::part(label))

Parts are the escape hatch.

Lets the consumers tweak layout or typography without us having to add new tokens or variants for every edge case.

### Example:

```
w-radio::part(label) {
  font-weight: 600;
}

```
```
w-radio::part(control) {
  margin-top: 1px;
}

```

### Pros

- Great DX for small, local tweaks
- Avoids variant explosion
- No need to change the component API for one-offs

### Cons

- Less stable than tokens (structural contract)
- Easy to misuse for full theming
- Not meant for state logic

### Use for

- Spacing / alignment tweaks
- Minor typography adjustments
- One-off context-specific styling 

## Internal tokens (--_*)

These are implementation details.

### Rules:

- Defined in the base mapping (:host)
- Consumed by selectors only
- Never overridden from the outside

### Example:

```
[part='control'] {
  border-color: var(--_border-color);
  background-color: var(--_bg);
}
```

This is what lets us change selectors or structure later without breaking consumers.

## Recommended order of use

### Rough rule of thumb:
- Do nothing → component follows theme defaults
- Override --w-c-* → supported, scalable customization
- Use ::part(...) → last-mile, local tweaks

If it feels like you need to override internal tokens or repeat selectors, that’s usually a sign we’re missing either:

a --w-c-* token, or a part

## Why this setup

- Should scale for many components and contributors
- Keeps styling declarative and CSS-only
- Gives consumers flexibility without locking us into internals

PR #506, opened 2026-01-30
chore(build): add backwards compatibility path validation

digitalsadhu:
make sure a recent issue doesn't happen again

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install Playwright
         run: pnpm exec playwright install --with-deps
       - name: Test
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Install Playwright
         run: pnpm exec playwright install --with-deps
       - name: Test
-        run: pnpm test && pnpm test:module-resolution && pnpm test:output-validation
+        run: pnpm test && pnpm test:module-resolution && pnpm test:output-validation && pnpm test:validate-eik-backwards-compat.js
```
imprashast: Do you want to run this on cli on every release? this includes every next release too?

PR #511, opened 2026-02-02
chore(test): add wcag test setup

digitalsadhu:
Testing out adding WCAG automated testing to the textarea

I reached out out to the accessibility team to see what they thought. Tom was positive to adding WCAG tests.

### DIFF
--- a/setup-tests.ts
+++ b/setup-tests.ts
@@ -5,7 +5,6 @@ import type { RenderResult } from 'vitest-browser-lit';
 import { expect } from 'vitest';
 
 interface AxeMatchers {
-  toHaveNoViolations(): void;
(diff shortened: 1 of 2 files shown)


setup-tests.ts
```diff
@@ -1 +1,56 @@
 import 'vitest-browser-lit';
+import { i18n } from '@lingui/core';
+import { expect } from 'vitest';
+import type { AxeResults, Result } from 'axe-core';
+
+// Initialize i18n with English locale for tests
+i18n.load('en', {});
```
wkillerud: This feels like something that could make future tests of accessible labels harder.

It's perhaps more tedious, but I would expect component tests to load and activate the component's English locale, if any.

PR #512, opened 2026-02-03
chore: add CSS parts to components

pearofducks:
I was trying to style Expandable's button today via `button-class` and it doesn't work anymore (I think due to being wrapped by the unstyled-heading element).

I added `part` to Expandable and then got AI to do the same for other components.

If the formatting the bot applied needs reverted I'm happy to go through and do that, overall it seems to align with some of the less-terse styling I've seen applied elsewhere though.

### DIFF
--- a/packages/alert/alert.ts
+++ b/packages/alert/alert.ts
@@ -123,9 +123,9 @@ class WarpAlert extends LitElement {
   render() {
     return html`
       <w-expand-transition ?show=${this.show}>
-        <div role=${this.role} class=${this._wrapperClasses}>
-          <div class=${this._iconClasses}>${this._icon}</div>
-          <div class=${ccAlert.textWrapper}>
+        <div role=${this.role} class=${this._wrapperClasses} part="alert">
+          <div class=${this._iconClasses} part="alert-icon">${this._icon}</div>
+          <div class=${ccAlert.textWrapper} part="alert-content">
             <slot></slot>
           </div>
         </div>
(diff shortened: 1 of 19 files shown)


digitalsadhu:
I had a convo with @Skadefryd the other day about the use of parts. We agreed to delay so that he could have time to have a think about how we should apply them strategically so it be great if he can chip in here. 

I have a few thoughts: 

Parts are part of the public facing API surface of components so.. where does it make sense to apply them?

Once applied, taking them away again becomes a breaking change and worth some thought before we do. 

If we apply parts too liberally, we risk essentially bleeding our internal implementation detail out and making it hard to refactor the internals without breaking the part api surface. If we can design/abstract the parts in a way that doesn't couple them too heavily with the implementation details, that would be good.

Also, we have other dials to use as well as (or instead of parts). CSS variables and reflected attributes

**CSS variables** 
- We should probably use these rather than parts for any specific targeted configurability eg to make corner radius configurable or drop shadows or what have you.

**CSS variables with reflected attributes**
- If we hang our styles off attributes (eg. invalid) I think we can target variables for different states.

eg.

internal:

```css
:host { -border-opacity: 0.5; }
```

external override with

```css
my-button[disabled] { --border-opacity: 0.9; }
```

digitalsadhu:
Great plan! Let's start with expandable. Just had a chat with @imprashast and @Skadefryd around the water cooler (so to speak) and I get the impression that starting with expandable, nail some conventions/approaches together would be a good way to go.

PR #519, opened 2026-02-11
fix: align sliders with updated design spec

wkillerud:
- Add example for visually hiding the textfield
- Add example for hiding the minimum and maximum labels
- Update open ended labels and values to match spec (Før 1950, 2025+ and Min, Max)
- `formatter` into `valueFormatter` and `labelFormatter` since they will be different in some cases
- Fix disabled state CSS
- Suggested improvements in screen reader announcement
- Fix tab order in range sliders (thumbs first, then input fields)
- Show validation errors on blur, not right away

Docs PR: https://github.com/warp-ds/docs/pull/190

### DIFF
--- a/packages/slider-thumb/slider-thumb.ts
+++ b/packages/slider-thumb/slider-thumb.ts
@@ -34,56 +34,55 @@ class WarpSliderThumb extends FormControlMixin(LitElement) {
   ariaDescription: string;
 
   @property({ reflect: true })
-  label: string;
-
-  @property({ reflect: true })
   name: string;
 
   @property({ reflect: true })
   value: string;
 
+  /** @internal Set by `<w-slider>` */
   @property({ type: Boolean, reflect: true })
   disabled: boolean;
 
+  /** @internal Set by `<w-slider>` */
   @property({ type: Boolean, reflect: true })
   invalid = false;
 
-  /** Set by `<w-slider>` */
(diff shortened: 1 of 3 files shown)


packages/slider-thumb/styles/w-slider-thumb.styles.ts
```diff
@@ -72,9 +76,13 @@ export const wSliderThumbStyles = css`
     box-shadow: none;
   }
 
+  .w-slider-thumb__range[disabled]::-moz-range-thumb{
```
pearofducks:
```suggestion
  .w-slider-thumb__range[disabled]::-moz-range-thumb {
```

wkillerud:
We have a known issue in Firefox where anchor positioning exists, so the polyfill doesn't get used, [but it's broken](https://bugzilla.mozilla.org/show_bug.cgi?id=2006551)...

packages/textfield/styles/w-textfield.styles.ts
```diff
@@ -36,6 +36,12 @@ export const wTextfieldStyles = css`
     display: none;
   }
 
+
+  /* It's supposed to behave like a placeholder, but look like a value. Don't tell the designers 🤫 */
```
digitalsadhu: shhhh

PR #520, opened 2026-02-13
feat: release checkbox and slider

digitalsadhu:


joms:
> > Remember to not squash merge
> 
> 😆 - so HAAAAANDS can forever be a commit in this repo

Don't forget "fix: it"

digitalsadhu:
🤦 I preferred squashing but @imprashast is right, it causes unnecessary faff. Your silly commits will be forever immortalized.

PR #525, opened 2026-02-17
feat: add radio button and group

digitalsadhu:
What this PR does:
* Adds the radio and radio-group components ready for release
* Removes the use of Web Awesome in the Radio by rewriting to add our own implementation for form participation, validation etc.
* Adds lots of tests
* Adds A11y tests
* Refactors and improves the radio-group to match Figma designs. group label and help text, validation for the group, form participation for the group etc.
* Adds a styling API (CSS custom variables and parts) to the radio
* Adds support for using the radio on its own like a native radio
* Adds lots of stories

Also lots of testing and fixing. Best effort testing with VoiceOver seems good to me (but I barely know what Im doing).

### DIFF
--- a/packages/radio-group/radio-group.ts
+++ b/packages/radio-group/radio-group.ts
@@ -176,7 +176,7 @@ export class WRadioGroup extends FormControlMixin(LitElement) {
   }
 
   private handleInvalid = (event: Event) => {
-    // Prevent native validation popover. We'll surface errors via help text.
+    // Prevent native validation popover. We surface errors via help text.
     event.preventDefault();
     this.hasInteracted = true;
     this.updateValidity();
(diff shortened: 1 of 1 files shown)


wkillerud (commented):
The focus outline is a bit different between radio and checkbox. The radio highlights both the circle and label, while checkbox only highlights the label.

There's some inconsistency in the radio examples as well, but that might just be storybook. Some examples highlight the full width of the example container on focus while others wrap around the radio + label.

Otherwise looks good to me, nice work!

digitalsadhu:
> The focus outline is a bit different between radio and checkbox. The radio highlights both the circle and label, while checkbox only highlights the label.
> 
> There's some inconsistency in the radio examples as well, but that might just be storybook. Some examples highlight the full width of the example container on focus while others wrap around the radio + label.
> 
> Otherwise looks good to me, nice work!

thanks! ill check it out

Issue #527, opened 2026-02-23
Storybook examples should be localized


wkillerud:
To avoid confusion when testing, strings we use in Storybook stories should be localized, same as strings we use in components. That way we don't get some strings in English and some in Norwegian, for example.

Make sure they don't ship as part of the final bundle though.

wkillerud:
We might also want to have separate controls in Storybook for language, as opposed to tying it to brand like we do now.

digitalsadhu:
Quick fix applied in #540 
Keeping the issue around as we think it might be nice to be able to switch languages in the examples too.

PR #540, opened 2026-02-26
docs: storybook localization

wkillerud:
Taking the shortcut for now, English only.

Relates to #527, but we should perhaps keep the issue open if we want to support translated demos?


imprashast (approved):

digitalsadhu (approved):

digitalsadhu:
Lets make a note on the issue about it

PR #543, opened 2026-02-26
fix: hover style for datepicker

### LINKED ISSUE #301: Create beta button (very optional)
Implement using the version in @warp-ds/components as a guide and remember form associated stuff from the @warp-ds/elements version.

### LINKED ISSUE #537: Calendar button misalignment
Calendar icon button on the right is out by a couple px

Image

### LINKED ISSUE #541: w-link is not focusable with a keyboard in Vivaldi
- Go to https://warp-ds.github.io/elements/?path=/docs/buttons-link--docs
- Tab through the examples.
- Only the Show code button gets keyboard focus.

Vivaldi
Versjon: 7.8.3925.74 (Offisiell versjon) (arm64)
Chromium versjon: 144.0.7559.230
Kanal: Offisiell versjon
Plattform / OS: macOS Version 26.3 (Build 25D125)

wkillerud:
the famous scene from Malcolm in the Middle where Hal keeps finding new things to fix when all he wanted to do was change the lightbulb

The `w-link` change is debatable. I can reproduce the bug in a clean version of the Vivaldi browser (Guest profile). It's not a problem in Chrome, Safari or Firefox.

BEMifying the button so I can adjust the things I need to make the button properly fit inside the input field. I tested the Storybook examples where we use `w-button`. Looks right to me.

The button changes could use an extra pair of eyes and some QA on Storybook 🙏 

Closes #537
Closes #541


packages/button/button.ts
```diff
@@ -201,13 +118,6 @@ class WarpButton extends FormControlMixin(LitElement) {
   // capture the initial value using connectedCallback and #initialValue
   #initialValue: string | null = null;
 
-  static styles = [
-    reset,
-    styles,
-    css`:host([full-width]) { width: 100%; }`,
-    css`.rounded-radius-default{border-radius:var(--w-button-radius-default,8px);}.rounded-radius-utility{border-radius:var(--w-button-radius-utility,4px);}`,
```
imprashast, line 208: One thing I immediately found was that these tokens were missing from the new stylesheet. 
To see the change, you can switch to vend-light theme in next and compare

packages/button/button.ts
```diff
@@ -329,27 +164,71 @@ class WarpButton extends FormControlMixin(LitElement) {
...
+            class=${this.buttonClass}
+            rel=${this.target === '_blank' ? this.rel || 'noopener' : undefined}
+          >
+            <slot></slot>
+          </w-link>
+        `
+        : html`
+          <button
+            type=${this.type || 'button'}
+            class=${classMap({
```
imprashast: Not setting `buttonClass` anywhere in this?

packages/button/button.ts
```diff
@@ -329,27 +164,71 @@ class WarpButton extends FormControlMixin(LitElement) {
...
+              'w-button--quiet':
+                // only include `quiet` with the backwards compat if the variant does not have its own -quiet modifier.
+                (this.quiet && !['negative', 'utility', 'overlay', 'overlayInverted'].includes(this.variant)) ||
+                this.variant === 'quiet',
+              'w-button--negative-quiet':
+                (this.quiet && this.variant === 'negative') || this.variant === 'negativeQuiet',
+              'w-button--utility-quiet': (this.quiet && this.variant === 'utility') || this.variant === 'utilityQuiet',
+              'w-button--overlay': !this.quiet && this.variant === 'overlay',
+              'w-button--overlay-inverted': !this.quiet && this.variant === 'overlayInverted',
+              'w-button--overlay-quiet': (this.quiet && this.variant === 'overlay') || this.variant === 'overlayQuiet',
```
digitalsadhu: could all these be moved into the CSS?
```css
:host([quiet][variant="overlay"]) a {}
```

packages/button/button.ts
```diff
@@ -15,139 +13,49 @@ import { messages as enMessages } from './locales/en/messages.mjs';
...
+  'utilityQuiet',
+  'overlay',
+  'overlayInverted',
+  'overlayQuiet',
+  'overlayInvertedQuiet',
+];
 
 type ButtonType = 'button' | 'submit' | 'reset';
 
-const buttonDefaultStyling = 'font-bold focusable justify-center transition-colors ease-in-out';
```
imprashast, line 26: Good to get rid of this mess 🥇

packages/button/button.ts
```diff
@@ -180,12 +91,18 @@ class WarpButton extends FormControlMixin(LitElement) {
...
+  disabled: boolean;
+
   @property({ reflect: true })
   rel: string;
 
   @property({ attribute: 'full-width', type: Boolean, reflect: true })
   fullWidth: boolean;
 
+  /**
+   * @deprecated This class is applied inside the shadow DOM and is unlikely to have the desired effect. Use attributes or CSS variables to customize the appearance of the button.
```
digitalsadhu, line 105: It would be nice to ship a :parts and CSS custom properties API with this if possible in order to have a proper alternative to using this button-class attribute. @Skadefryd has a plan for this so maybe now is a good time to proritize button as part of this PR?

digitalsadhu (commented):
Nice! Be great if @Skadefryd can take a look.

packages/button/button.ts
wkillerud: Ooh, _that's_ what it was for, oops! Good catch, I'll get it back in working order.

packages/button/button.ts
wkillerud: Oh right, that can probably be a bit cleaner. Is there a consensus on this style, attribute selectors like that over CSS classes? (I'm a fan, just asking)

digitalsadhu: As far as I know, we are all fans. I think.

packages/button/button.ts
wkillerud: I'll leave that to #579

packages/link/link.ts
```diff
@@ -46,19 +46,19 @@ class WarpLink extends LitElement {
   };
 
   @property({ type: Boolean, reflect: true })
-  autofocus = false;
```
digitalsadhu, line 49: why?

digitalsadhu (approved):

packages/link/link.ts
wkillerud: Didn't this type of code cause a hydration problem?

PR #557, opened 2026-03-03
fix(switch): properly enable keyboard navigation and labelling

digitalsadhu:
Switch labelling and keyboard navigation weren't working. This PR addresses that.

### DIFF
--- a/packages/switch/switch.ts
+++ b/packages/switch/switch.ts
@@ -2,7 +2,7 @@
 
 import { classNames } from '@chbphone55/classnames';
 import { FormControlMixin } from '@open-wc/form-control';
-import { html, LitElement } from 'lit';
+import { css, html, LitElement } from 'lit';
 import { property } from 'lit/decorators.js';
 
 import { reset } from '../styles';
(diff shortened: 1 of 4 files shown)


packages/switch/switch.ts
```diff
@@ -127,10 +171,9 @@ export class WarpSwitch extends FormControlMixin(LitElement) {
       <div>
         <button
           type="button"
-          role="switch"
-          aria-checked=${this.checked}
+          role="none"
```
pearofducks, line 197: This is the element that has `focusable`. Is that going to be a problem now that it's got `role=none` and `tabindex=-1`?

(Or should it stop having focusable so we can clearly see which element owns that UI?)

joms (approved):
Image

packages/switch/switch.ts
digitalsadhu: Yea after looking at this with Martin we decided that since relationships (in this case label and switch) need to be in the same dom (either light dom or same shadow dom), its best to move all ARIA attrs to the host and hide the internal details from screen readers. Focusable is a good point, ill take a look at it but I guess is should not be on the button

digitalsadhu:
You were quite right @pearofducks focus ring was borked. Fixed in latest commit. Great catch thanks!

PR #567, opened 2026-03-19
fix: prevent hydration mismatch errors by using element internals

digitalsadhu:


digitalsadhu:
Tested and Ok'd by Martin

PR #568, opened 2026-03-20
fix(select): preserve framework-set value in firstUpdated

turkus:
WARP's firstUpdated calls #syncFromNativeSelect which reads the native <select> value (browser defaults to 1st option) before the native select has been synced with the host's value property. This overwrites framework-set values (e.g. React's value prop). Fix: sync native select with host value before reconciling.


digitalsadhu (approved):
Nice thanks!

Issue #566, opened 2026-03-18
Reflecting attributes and hydration issues


digitalsadhu:
I may have discovered a couple interesting things regarding custom elements and a11y.

* setting attributes on the host from inside the component causes hydration warnings in react router/remix/next.js apps. These warnings are only warnings and can safely be suppressed but are very noisy
* It seems that the elementsInternals API may have a solution in that you can set aria attributes on this object internally in the element and the browser should respect this info without the need to set the aria attributes explicitly on the host itself (untested by me yet)

Eg.

```css
class WButton extends HTMLElement {
    constructor() {
      super();
      this.internals = this.attachInternals();
      this.internals.role = 'button';
      // Also available: ariaLabel, ariaDisabled, ariaPressed, etc.
    }
  }
```

* Setting default values where reflect is true will also cause these issues. Eg the button sets type and variant if the user doesn't set them. Styles are often hung off these attributes so they need to be present but this can be solved in the CSS 

Eg.

```css
:host(:not([variant])),
  :host([variant="primary"]) {
    /* primary styles */
  }
```

digitalsadhu:
began fixing this in #567

PR #571, opened 2026-03-24
fix: prevent hydration mismatches in React caused by client side DOM changes

digitalsadhu:
See #566 

This PR continues the work done in #567 

In this PR, reflected default values are removed and calculated as needed instead. This means that if an optional attribute is NOT provided, the default is applied without reflecting the attribute back into the DOM (which is what causes mismatch issues). None of this should be breaking since if a user was setting a value for a given attribute, nothing changes and if they weren't, appearance/behaviour should be the same but the attribute won't be present in the DOM. (though technically it could be possible that someone targetted said attribute with CSS or JS code which could cause an issue)

Second, in this PR, in a couple cases, Ive also changed the parent/child relationship where one exists AND the parent manipulates the child such that its not manipulating attributes (and therefore changing the DOM) and instead manipulates properties.

For example:
Before this change the hidden attribute would be set by w-tab on w-tab-panel. After the change, it is set as a property instead and so does not change the DOM.
```html
<w-tab> 
  <w-tab-panel hidden></w-tab-panel>
<w-tab>
```

Again, changes like this should continue to work as before but theres a small chance that someone could have used styling to target attributes that are only available as properties.

I think we should risk this change as a non breaking change even though technically there are edge cases that are breaking.

Ive done some digging with code search and found the following:
- w-tabs: not yet in use
- w-tab-panel: not yet in use
- w-alert: affects no one
- w-badge: affects no one
- w-button: affects no one
- w-combobox: not yet in use
- w-dead-toggle: not yet released
- w-icon: affects no one
- w-link: affects no one
- w-page-indicator: not yet in use
- w-pagination: not yet in use
- w-slider: affects no one
- w-slider-thumb: affects no one
- w-textfield: affects no one
- w-toast: affects no one

So in my mind we are good to cheat a little and publish these small breaking changes as non breaking

### DIFF
--- a/packages/alert/alert.test.ts
+++ b/packages/alert/alert.test.ts
@@ -19,3 +19,31 @@ test('does not render when "show" is not set / or not true', async () => {
   const page = render(component);
   await expect.element(page.container.querySelector('w-alert')).not.toHaveAttribute('show');
 });
+
+test('defaults to info variant when no variant attribute is set', async () => {
+  const component = html`<w-alert show data-testid="alert">Default alert</w-alert>`;
+  const page = render(component);
+
+  const el = page.getByTestId('alert').element() as HTMLElement;
+
+  // The variant attribute should not be reflected (to avoid hydration mismatch)
+  expect(el.hasAttribute('variant')).toBe(false);
+
+  // But the info icon should be rendered (default behavior)
+  await expect
+    .poll(() => {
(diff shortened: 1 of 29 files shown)


wkillerud (approved):
Makes sense, and I agree we can sneak this one in as non-breaking with the checks you've done 👍

PR #574, opened 2026-03-27
fix: refactor components to avoid React hydration mismatch errors

digitalsadhu:
Adds hydration mismatch tests and then attempts to fix using the strategy of

* a11y attrs set as properties on element internals, not on the host
* delegates focus onto tabindex onto an element in the shadowdom for enabling disabling tabbability of the component
* default attributes don't reflect default state onto the host, they just calculate defaults on the fly as needed

Breadcrumbs is still a problem. Also, I still need to go through everything here for another sweep before im satisfied we've nailed this.

### DIFF
--- a/packages/affix/affix.hydration.test.ts
+++ b/packages/affix/affix.hydration.test.ts
@@ -9,6 +9,11 @@ describe('w-affix React SSR hydration', () => {
     window.__HYDRATION_WARNINGS__ = [];
   });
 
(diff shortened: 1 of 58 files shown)


trygve-lie (approved):
Looks according to what we discussed 🥇

trygve-lie (approved):
lgtm 👍

PR #577, opened 2026-04-08
fix(combobox): hydration error fixes

digitalsadhu:
* Fixed combobox hydration issues in React 19 SSR flows by preventing unintended initial reflected attributes and hardening property handling during upgrade.
* Added targeted hydration coverage for the real-world dangerouslySetInnerHTML mismatch case and converted it from a known-failing repro into a passing regression test.
* Improved combobox property consistency by standardizing reflected/defaulted props and applying useDefault: true where initial reflection should be suppressed.

### DIFF
--- a/packages/combobox/combobox.ts
+++ b/packages/combobox/combobox.ts
@@ -51,31 +51,31 @@ export class WarpCombobox extends FormControlMixin(LitElement) {
   options: ComboboxOption[] = [];
 
   /** Label above input */
-  @property({ type: String, reflect: true })
-  label?: string;
+  @property({ type: String, reflect: true, useDefault: true })
+  label?: string = '';
 
   /** Input placeholder */
-  @property({ type: String, reflect: true })
-  placeholder?: string;
+  @property({ type: String, reflect: true, useDefault: true })
+  placeholder?: string = '';
 
   /** The input value */
-  @property({ type: String, reflect: true })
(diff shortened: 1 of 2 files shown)


pearofducks (approved):

packages/combobox/combobox.ts
```diff
@@ -99,12 +99,12 @@ export class WarpCombobox extends FormControlMixin(LitElement) {
...
-  @property({ type: String, reflect: true })
-  name?: string;
+  @property({ type: String, reflect: true, useDefault: true })
+  name?: string = '';
 
   /** Autocomplete attribute for the input field */
-  @property({ type: String, reflect: true })
-  autocomplete?: string;
+  @property({ type: String, reflect: true, useDefault: true })
+  autocomplete?: string = 'off';
```
pearofducks, line 107: Does this represent a behavior change for the user?

`undefined` would yield browser-default -> 'on' and now it's explicitly set to `'off'`

digitalsadhu: this used to be off, then we moved to defaulting it to off further down in the code to avoid reflection but now that I know about useDefault, I was able to put things back.

trygve-lie (approved):
🥇

packages/combobox/combobox.ts
pearofducks: No sweat just checking! Thanks!

Issue #524, opened 2026-02-16
Range slider (again)


joms:
Bugs that we have haven't yet discovered: 
- Error from "the outside" triggers immediately
- If you press arrow up/down the thumb value is set to `0 +/- 1` in stead of `min +/- 1` or `max +/- 1`.

Sidenote: There's some weird interaction when setting an invalid value. For instance when setting `min` to 2002, and `max` to 2000, both fields ends up at 2002. I might not just be handling it correct in my implementation though...
https://warp-ds.github.io/elements/?path=/story/forms-slider-and-range-slider--open-ended

<img width="627" height="467" alt="Image" src="https://github.com/user-attachments/assets/2fe7b9eb-9181-42c5-954b-2f77cc0d2fe5" />

Note that the search filter (see the pill `2002 - 2002`) is set to 2002 - 2002

joms:
Considering we are in a form element, I think we should run validation on keydown enter too, preventing submit if validation fails.

PR #585, opened 2026-04-16
fix(slider): bug where external validator would fire immediately

joms:
https://github.com/warp-ds/elements/issues/524

### DIFF
--- a/packages/slider/slider.test.ts
+++ b/packages/slider/slider.test.ts
@@ -449,7 +449,7 @@ test('disabled fieldset communicates disabled state', async () => {
 });
 
 // Error state accessibility (WCAG 3.3.1 Error Identification)
-test('invalid slider sets aria-invalid on fieldset', async () => {
+test('invalid slider does not set aria-invalid on fieldset before interaction', async () => {
   const component = html`
     <w-slider label="Invalid slider" min="0" max="100" invalid error="Please select a value">
       <w-slider-thumb name="value"></w-slider-thumb>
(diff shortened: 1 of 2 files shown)


packages/slider/slider.test.ts
```diff
@@ -460,15 +460,55 @@ test('invalid slider sets aria-invalid on fieldset', async () => {
...
+
+  const fieldset = slider.shadowRoot.querySelector('fieldset');
+
+  // Error should not show before interaction
+  expect(fieldset.getAttribute('aria-invalid')).toBeNull();
+});
+
+test('invalid slider sets aria-invalid on fieldset after blur', async () => {
+  const component = html`
+    <w-slider label="Invalid slider" min="0" max="100" invalid error="Please select a value">
```
wkillerud, line 472: I'm not 100% sold on this. Would you not expect to see a validation error as a developer when you set these attributes? Isn't it on the component user here to time this correctly?

What if I do a form submit, the server validates with an error and includes it in the markup. Now I won't see that error until I happen to interact with the slider. I might submit again, get it rejected, and have no visual explanation of why.

joms: Good point. This is following the pattern and the specification of not invalidating a field immediately though. So this fix is to get the external validation to have the same timing as the internal, which is based on the design specification.

Or did I misunderstand your comment? 🤔

joms: Or are you arguing it should technically be invalid, but not visually?

wkillerud: No, I'm arguing it should be visually (and technically) invalid when you set the attributes `invalid error="Some error message"`.

wkillerud: > This is following the pattern and the specification of not invalidating a field immediately though.

That makes sense on a first interaction, but in that case I would expect component users _not_ to set `invalid error="Some error message"`.

joms: Right. So not validating immediately until it has its first error? Sorry, I need it fead by teaspoon today. Trying to wrap my head around the different steps and its flow 😅

wkillerud: I claim this is user error 😄 "Error from "the outside" triggers immediately" is the way it should work.

joms:
We had a huddle where we found that William indeed was correct :(

PR #586, opened 2026-04-17
fix: use correct initial value and update active track on reset

wkillerud:
The native form reset didn't work as expected.


joms (approved):
Amaze 👏

PR #609, opened 2026-04-23
fix(attention): a11y improvements

digitalsadhu:
After a discussion with @storsletten, couple small improvements (even though we are going to deprecate this component and split it out into popover, callout and tooltip

* In the implementation, change the wrapping div to a section
* Remove the id/aria-details stuff


wkillerud (approved):

wkillerud:
Slider hydration thing that fails, let's just merge

PR #613, opened 2026-04-24
docs: build docs md files

digitalsadhu:
Setup the script to build docs files in dist/docs and then fleshed out alert as an example

We end up with colocated markdown files built mostly out of jsdoc and a couple .md files per component
These are then exported via package.json exports and can be consumed in the Warp docs site. If we decide to change the docs site later, this setup should work nicely for that too.

```
/packages/alert/docs
  - usage.md
  - accessibility.md
  - examples.md (which can later be used for development as a replacement for storybook
```

running build/docs.js will produce the following files in the dist folder
```
/dist/docs
  /alert
    - usage.md
    - accessibility.md
    - examples.md 
    - api.md
    - alert.md (full build into a single document for the component)
```

We could do more with jsdoc if we preferred. Eg. We could include accessibility and usage notes/docs in jsdoc and build them out of there (via CEM)


wkillerud (approved):
Very nice :shipit:

PR #615, opened 2026-04-27
refactor(button,link): migrate to component tokens styling API

Skadefryd:
- Replace UnoCSS classes with component tokens (--w-c-button-*)
- Fix semantic token bypasses in overlay variants
- Add styling-api.md documenting customization points (Goes into the docs for this component when ready)
- Add comprehensive test stories (AllVariantsGrid, QuietBackCompatMapping, LoadingAcrossVariants)
- Maintain backwards compatibility with quiet flag mapping
- Link now has missing hover/active states for overlay variants


packages/button/styling-api.md
```diff
@@ -0,0 +1,78 @@
+# w-button styling API
+
+This component supports styling through **component tokens** (CSS custom properties with a --w-c- prefix) and **parts**.
+
+---
+
+## Parts
+
+Use `::part(...)` from outside the component.
```
wkillerud: I wonder if the documentation of the parts and CSS properties can be moved to JSDoc so it can be included in the custom element manifest?

- https://custom-elements-manifest.open-wc.org/analyzer/getting-started/#documenting-your-components (`@csspart` and `@cssprop` in particular)

@digitalsadhu and I are working on a documentation build pipeline that autogenerates certain documentation from the custom element manifest, it would be sweet to get parts and CSS tokens in as part of that. Plus, it can be shown inline in code editors.

wkillerud (commented):
Cool stuff! The CSS API itself looks good to me ✅ 

I think `styles.ts` can get overwritten by automation. For the components that have non-Uno CSS there should be a `styles/` directory with a `<component name>.styles.ts` file. We can change that convention if you'd like, but it should probably be a different file from `styles.ts` in any case.

packages/button/styling-api.md
Skadefryd: YEs the thought of this file is that it should be included/moved to the docs for this element, its ment as a temporary thing its not forgotten. Would be ioce actually if the docs for each componernt lived closer to the code itself but i guess its getting less relevant with the tools we have now :)

digitalsadhu: Styles won't be built if `// @warp-css;` is removed from the top of any component file so its fine to call it styles.ts.

Agree with William, the docs you've added are perfect for adding into the new docs build pipeline. I don't mind if its via jsdoc or a separate .md file like this

digitalsadhu (approved):
Changes look good. Not sure why theres so much formatting noise? Are we bouncing back and forth between different settings?

wkillerud (approved):

PR #619, opened 2026-04-27
feat(affix): add icon attribute, automatic aria labels for search and clear, textfield reset and form submit

digitalsadhu:


packages/affix/docs/usage.md
```diff
@@ -1 +1,41 @@
...
+
+Use `clear` to render a clickable clear/reset icon button.
+
+```html
+<w-affix clear aria-label="Clear input"></w-affix>
+```
+
+### In Textfield Slots
+
+`w-affix` is usually slotted into the w-textfield form element. Use the textfield's `prefix` or `suffix` slots to add the affix to the component.
```
wkillerud: Maybe we can move this example and text up to under the `Usage` heading? Then the rest of the examples make more sense on their own.

wkillerud (approved):
Small suggestion, up to you

packages/affix/affix.ts
```diff
@@ -1,15 +1,18 @@
...
 import { property } from 'lit/decorators.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
+import { FormControlMixin } from '@open-wc/form-control';
 
 import { reset } from '../styles.js';
 
 import { styles } from './styles.js';
 import '../icon/icon.js';
 import { detectLocale } from '../i18n.js';
+import { WarpTextField } from '../textfield/textfield.js';
```
wkillerud: guess it doesn't matter too much, but this can make it clearer we just use it for types

```suggestion
import type { WarpTextField } from '../textfield/textfield.js';
```

wkillerud (approved):

PR #620, opened 2026-04-27
docs: colocate toast docs

wkillerud:
Toast is special, so adapt the docs build pipeline ever so slightly.


packages/toast/docs/api.md
```diff
@@ -0,0 +1,71 @@
...
+| --- | --- | --- | --- |
+| id | `string` | Random | Random generated unique ID for the toast element |
+| type | [`ToastVariants`](#toastvariants) | `'success'` | Visual style of the toast |
+| text | `string` | `-` | The given toast message |
+| duration | `number` | `5000` | Duration of toast in milliseconds |
+| canclose | `boolean` | `false` | See [Dismissable toast](#dismissable-toast) |
+
+#### ToastVariants
+
+`'success' | 'warning' | 'error'`
```
digitalsadhu, line 71: we needed to add this file manually as it couldn't be generated from CEM?

digitalsadhu (approved):

packages/toast/docs/api.md
wkillerud: Yes, the API isn't a custom element, so no CEM

PR #621, opened 2026-04-27
chore: set up a simple watcher to rebuild docs on MD changes

wkillerud:
Rebuilds the CEM and docs in the `dist/` folder when we change `.ts` and `.md` files, for maximum developer velocity.

`watch:npm` doesn't work, assume it won't be missed.


digitalsadhu (approved):
Whooooa that snail is fast! 

https://youtu.be/vZopX05wXH0

PR #623, opened 2026-04-29
fix(datepicker): input field not updating after typing

BalbinaK:
## Summary
Fixes three input handling issues in the datepicker:

1. Popover selection not updating input after typing - the input used an attribute binding (`value=`) which stops reflecting changes once the user has typed. Switched to a property binding (`.value=`).
2. Change event not dispatched after typing - the native change event doesn't cross the shadow DOM boundary. Now suppressed and re-dispatched from the host element on blur.
3. Date segments resetting while editing - typing in one segment (e.g. day in dd/mm/yyyy) caused all segments to clear because the intermediate empty value triggered a re-render. Input now only syncs complete dates - final value is committed on blur.

## Demo

https://github.com/user-attachments/assets/06f9b70c-c209-4e3c-9a88-a1b14da75921

### DIFF
--- a/packages/datepicker/datepicker.ts
+++ b/packages/datepicker/datepicker.ts
@@ -183,8 +183,8 @@ class WarpDatepicker extends FormControlMixin(LitElement) {
     return startOfMonth(this.navigationDate);
   }
 
-  // capture the initial value using connectedCallback and #initialValue
   #initialValue: string | null = null;
+  #lastDispatchedValue: string | null = null;
 
   @state()
   get weeks() {
@@ -269,10 +269,9 @@ class WarpDatepicker extends FormControlMixin(LitElement) {
   }
 
   async #dispatchChangeEvent() {
-    // Let Lit finish rendering the updated value for the input field so
-    // the `event.target.value` is correct.
     await this.updateComplete;
-    this.input.dispatchEvent(new Event('change', { bubbles: true, composed: true }));
+    this.#lastDispatchedValue = this.value;
+    this.dispatchEvent(new Event('change', { bubbles: true }));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


wkillerud:
Thanks for the PR!

> Popover selection not updating input after typing

~I can't reproduce this issue on [Storybook](https://warp-ds.github.io/elements/?path=/docs/forms-datepicker--docs), the input updates the calendar OK.~

I misunderstood the problem, I thought it was about the popover not updating as you typed, but it's about selecting in the popover after having typed that does not update the input value. Can reproduce 👍 

> Change event not dispatched after typing

I'm not sure that's the expected behavior for a date input. [According to MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event):

> Unlike the [input](https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event) event, the change event is not necessarily fired for each alteration to an element's value.
> [...]
> the change event fires at a different moment:
> [...]
>  - When the user commits the change explicitly (e.g., by selecting a value from a [<select>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/select)'s dropdown with a mouse click, by selecting a date from a date picker for [<input type="date">](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/date)

So I would argue the change event is technically working as intended, as in "like `input type="date"` does.

> Date segments resetting while editing 

I can't reproduce this either on [Storybook](https://warp-ds.github.io/elements/?path=/docs/forms-datepicker--docs) I'm afraid.


https://github.com/user-attachments/assets/deb256c6-de16-482c-8dbe-990c85c91772

wkillerud:
Summarizing some things we learned:

- If the user has changed the value of an input field, changing the `value` attribute has no effect. The users' input takes precedence visually and in the DOM.
- To fix that, update the property in the DOM, not just the attribute in HTML.
- We weren't forwarding `input` or `blur` events in the React wrapper.
- `blur` events fire after both selecting the date from the calendar and when the user is done typing, so we should probably recommend that as the default usage unless you have specific validation/timing needs. If you do, use the `input` event for typing, `change` event for selecting in the calendar (same as `type="date"`).

@wkillerud will try to make that a bit more easy to digest in the Warp docs 😅

BalbinaK:
> Summarizing some things we learned:
> 
> * If the user has changed the value of an input field, changing the `value` attribute has no effect. The users' input takes precedence visually and in the DOM.
> * To fix that, update the property in the DOM, not just the attribute in HTML.
> * We weren't forwarding `input` or `blur` events in the React wrapper.
> * `blur` events fire after both selecting the date from the calendar and when the user is done typing, so we should probably recommend that as the default usage unless you have specific validation/timing needs. If you do, use the `input` event for typing, `change` event for selecting in the calendar (same as `type="date"`).
> 
> @wkillerud will try to make that a bit more easy to digest in the Warp docs 😅

Thank you for the summary and coming up with an alternative solution! I decided to open a separate PR after all 😅: https://github.com/warp-ds/elements/pull/627

PR #630, opened 2026-04-30
fix: submit the associated form on Enter like native form elements

wkillerud:
Discovered this while writing https://github.com/warp-ds/docs/pull/213, wanted to test that I wasn't lying about `blur` and `submit`. Turns out we didn't submit on Enter 😅


wkillerud:
I was right about `blur` not firing on `submit` though!

digitalsadhu (approved):
Ooopss

PR #642, opened 2026-05-04
fix: don't prevent default unless there's something to prevent 

### LINKED ISSUE #641: Can't type in an open ended range slider field without a value
- Go to https://warp-ds.github.io/elements/?path=/story/forms-slider-and-range-slider--open-ended
- Click in the Min input field
- Try to type `1932`

Arrow up and down works, and after that I can edit. I can also edit if I put the slider anywhere other than the edges.

wkillerud:
I broked it. I fixded it.

Also colocate docs.

### DIFF
--- a/packages/slider-thumb/slider-thumb.ts
+++ b/packages/slider-thumb/slider-thumb.ts
@@ -390,7 +390,6 @@ class WarpSliderThumb extends FormControlMixin(LitElement) {
     // Handle the case where an open ended slider that has
     // no current value (Min or Max) and the user presses
     // the up or down arrow in the input field.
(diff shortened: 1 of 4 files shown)


digitalsadhu:
Chocolate docs?

joms:
Merged with next and fixed conflict, assuming @wkillerud's solution is better than mine :shrug:

joms (approved):

digitalsadhu (approved):
This is a far superior solution than Jo Emil's solution in every way. We should not use Jo Emil's solution because this solution is a better solution... than Jo Emil's solution

PR #646, opened 2026-05-08
fix(button): wrap react button wrapper and map camel case to snake case

digitalsadhu:
Fixes an issue reported regarding camel case property names in React wrappers. We will need to do this for all components that use any camel case properties.

### DIFF
--- a/packages/button/react.ts
+++ b/packages/button/react.ts
@@ -5,19 +5,7 @@ import React from 'react';
 import { WarpButton } from './button.js';
 
 // decouple from CDN by providing a dummy class
-class Component extends LitElement {
-  get fullWidth(): boolean {
-    return false;
-  }
-
-  set fullWidth(_value: boolean) {}
-
-  get iconOnly(): boolean {
-    return false;
-  }
-
-  set iconOnly(_value: boolean) {}
-}
+class Component extends LitElement {}
 
 const BaseButton = createComponent({
   tagName: 'w-button',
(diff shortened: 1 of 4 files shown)


packages/button/react.ts
```diff
@@ -7,8 +7,29 @@ import { WarpButton } from './button.js';
...
   tagName: 'w-button',
   elementClass: Component as unknown as typeof WarpButton,
   react: React,
 });
+
+type BaseButtonProps = React.ComponentPropsWithoutRef<typeof BaseButton>;
+
+type ButtonProps = Omit<BaseButtonProps, 'full-width' | 'icon-only'> & {
+  fullWidth?: boolean;
+  iconOnly?: boolean;
```
BalbinaK, line 20: FYI, I don't think this property is actually used for anything. At least that was a result of a quick search in the code.

packages/button/react.ts
```diff
@@ -7,8 +7,29 @@ import { WarpButton } from './button.js';
...
+};
+
+export const Button = React.forwardRef<WarpButton, ButtonProps>(({ fullWidth, iconOnly, ...props }, ref) =>
+  React.createElement(BaseButton, {
+    ...props,
+    ...(fullWidth ? { 'full-width': true } : {}),
```
wkillerud, line 26: Did you check how this component behaves client-side now? The Lit docs give me the impression that these React props always map to properties as opposed to attributes, so I wonder if we now break that relationship – that the wrapper will try to set `element["full-width"]` instead of `element.fullWidth` essentially.

https://lit.dev/docs/frameworks/react/#createcomponent

packages/button/react.ts
digitalsadhu: Thanks Balbina, I suspect thats the case too.

packages/button/react.ts
digitalsadhu: Hmm yea that tracks. Ill test some more.

digitalsadhu: Seems to work as expected. I think that makes sense though. The wrapper will re-render the custom element based on the property which always maps fullWidth to full-width. Right?

wkillerud (approved):

PR #654, opened 2026-05-11
docs(pagination): colocate docs with code

wkillerud:


packages/pagination/docs/examples.md
```diff
@@ -1 +1,21 @@
...
+<script type="module">
+  const pagination = document.querySelector('w-pagination');
+
+  pagination.addEventListener('page-click', (event) => {
+    event.preventDefault();
+    // event.detail = { clickedPage: string }
+    console.log('Do something with the clicked page after preventing default navigation:', event.detail.clickedPage);
+  });
+</script>
+```
```
digitalsadhu, line 21: how/where do we want to document usage in a React app for things like this where html is more verbose?

digitalsadhu (approved):

packages/pagination/docs/examples.md
wkillerud: I've been thinking we document how to use the React wrappers in a separate file if that's what you mean, and show that on f. ex. https://warp-ds.github.io/docs/components/pagination/frameworks/react#pagination-react

PR #657, opened 2026-05-11
fix(modal-header): prevent close and back buttons from inheriting modal title tooltip

joakim-sch:
## Description
The close and back buttons inside modal-header were inheriting the host element's title attribute (the modal heading text) as a tooltip on hover. This happened because the title property on the Lit component maps to the native HTML title attribute, which browsers propagate to all descendant elements.

### Changes
Extract the i18n label into a const in both closeButton and backButton getters to avoid duplication
Set title on both buttons to match their aria-label, so hovering shows "Close" / "Back" (localized) instead of the modal heading

### DIFF
--- a/packages/modal-header/modal-header.ts
+++ b/packages/modal-header/modal-header.ts
@@ -64,16 +64,15 @@ export class ModalHeader extends CanCloseMixin(LitElement) {
   }
 
   get backButton() {
-    const backLabel = i18n._({
-      id: 'modal.aria.back',
-      message: 'Back',
-      comment: 'Aria label for the back button in modal',
-    });
     return this.back && !this._hasTopContent // Not showing back button when there is a top image
       ? html`<button
           type="button"
-          title="${backLabel}"
-          aria-label="${backLabel}"
+          title=""
+          aria-label="${i18n._({
+            id: 'modal.aria.back',
+            message: 'Back',
+            comment: 'Aria label for the back button in modal',
+          })}"
           class="header-button header-button-left"
           @click="${this.emitBack}">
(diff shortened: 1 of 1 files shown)


packages/modal-header/modal-header.ts
```diff
@@ -64,14 +64,16 @@ export class ModalHeader extends CanCloseMixin(LitElement) {
...
+    });
     return this.back && !this._hasTopContent // Not showing back button when there is a top image
       ? html`<button
           type="button"
-          aria-label="${i18n._({
-            id: 'modal.aria.back',
-            message: 'Back',
-            comment: 'Aria label for the back button in modal',
-          })}"
+          title="${backLabel}"
```
digitalsadhu: I wonder if setting title to an empty string is actually the correct move here.

https://www.a11yproject.com/posts/title-attributes/#:~:text=Short%20answer%3A%20Don't%20use%20them%2C,but%20the%20opposite%20is%20true.

digitalsadhu (commented):
Good catch that the title inherits, I wasn't aware of that. Looking at a11y best practices, it's suggested we don't use the title attribute at all in which case the solution would be to just set it to an empty string to avoid the inheritance.

digitalsadhu:
Adding @storsletten as a reviewer to get his thoughts on the a11y side

packages/modal-header/modal-header.ts
joakim-sch: After reading up on the subject matter I agree. PR updated 👍🏼

digitalsadhu:
Chatted with @storsletten, titles=bad so setting them to empty strings was the right move.
Thanks for the contribution @joakim-sch

PR #660, opened 2026-05-12
docs(link): colocate docs with code

wkillerud:


packages/link/docs/usage.md
```diff
@@ -1 +1,20 @@
...
+
+**Do:**
+
+- Use the `<w-link>` component when you need a link that looks like a button.
+
+**Do Not:**
+
+- Use the `<w-link>` component when you need a link that looks like a link. In that case, use the native `<a>`tag.
+- Use the `<w-link>` component when you need a button that looks like a button. In that case, use the native `<w-button>`tag.
+- Use the `<w-link>` component when you need a button that looks like a link. In that case, use the native `<w-button>`tag.
```
digitalsadhu: I just dont think this is a use case that we should support. I know we currently do... I just dont think it's valid. Right?

digitalsadhu (commented):
LGTM other than the comment about buttons that look like links. Maybe we just omit talking about that case?

PR #662, opened 2026-05-12
docs(icon): colocate docs with code

wkillerud:


packages/icon/icon.ts
```diff
@@ -28,18 +28,26 @@ function cacheFetch<T = string>(uri: string, options: CacheFetchOptions<T> = {})
...
   name!: string;
 
-  /** Size: small, medium, large or pixel value (e.g. "32px") */
+  /** 
+   * Size: small, medium, large or pixel value (e.g. "32px").
+   * 
+   * @default "medium"
+   */
   @property({ type: String, reflect: true })
   size: 'small' | 'medium' | 'large' | string;
```
digitalsadhu, line 44: We can add a default if we set useDefault: true for the property then you wouldn't need the @default I guess?

packages/icon/docs/examples.md
```diff
@@ -0,0 +1,45 @@
...
+
+You can set a pixel size as well if none of the presets fit your needs.
+
+<elements-example>
+
+```html
+<w-icon name="Rocket" size="13.37px"></w-icon>
+```
+
+</elements-example>
```
digitalsadhu, line 45: should we include an interactive icons table in colocated here somewhere?

digitalsadhu (approved):

packages/icon/docs/examples.md
wkillerud: The source of truth for the list of icons [is in the Eikons repo](https://raw.githubusercontent.com/warp-ds/Eikons/refs/heads/main/icon-keywords.json). The docs overview [generates the list based on that JSON](https://github.com/warp-ds/docs/blob/main/docs/components/icons/iconNames.data.js). I don't really want to duplicate the list manually. We could add a similar build step of course that fetches the list of icons and inlines it.

digitalsadhu: Fine by me either way

wkillerud: Let's see how this current one does then, less duplication

PR #663, opened 2026-05-12
feat(combobox): add support for option element children

digitalsadhu:
This had been irking me for a while so decided to fix while I was looking at the docs.
Essentially mirrors how select works and allows option elements to be used as children to define options for the combobox.

```js
<w-combobox label="Select a fruit">
  <option value="apple">Apple</option>
  <option value="banana">Banana</option>
  <option value="orange">Orange</option>
</w-combobox>
```

Also still supports the .options property syntax

Also adds docs


wkillerud (approved):
Heck yeah 🙌

PR #665, opened 2026-05-13
feat(snackbar): add the snackbar component for Elements

wkillerud:
Opening a draft for some early feedback


packages/snackbar/docs/usage.md
```diff
@@ -0,0 +1,51 @@
...
+
+const item: WarpSnackbarItem = create('Snackbar message');
+```
+
+### `<w-snackbar>` JS API
+
+The `create` function in `@warp-ds/elements/snackbar` is a convenience wrapper around the `create` function on `<w-snackbar>` itself. In other words you can create snackbar items without importing anything.
+
+```js
+let snacbkar = document.querySelector("w-snackbar");
```
digitalsadhu: snoof gbar?
goodfrabar?

https://tenor.com/view/jack-nicholson-goos-fraba-woosah-bad-boys-gif-10675202

packages/snackbar/docs/usage.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+### `@warp-ds/elements` JS API
+
+You can import the `create` function from `@warp-ds/elements/snackbar`. It creates a snackbar item and adds it to the snackbar.
+
+```ts
+import { create, type WarpSnackbarItem } from "@warp-ds/elements/snackbar";
+
+const item: WarpSnackbarItem = create('Snackbar message');
```
digitalsadhu: Since this is just a thin wrapper, and if we ship <w-snackbar> into browsers, would it make sense to also ship the create function so that apps dont need to import anything and don't end up bundling the same snippet of code over and over? Eg. window.create. Maybe thats just too yuck to even consider. 
Maybe we should tell folks to use it via the element.create method...
Is there a declarative way to do this?

digitalsadhu: Web awesome does it (toast not snackbar in this case) with a create on the component
```html
<div id="toast-basic">
  <wa-button appearance="filled">Show notification</wa-button>
  <wa-toast></wa-toast>
</div>

<script>
  const container = document.getElementById('toast-basic');
  const toast = container.querySelector('wa-toast');
  const button = container.querySelector('wa-button');

  button.addEventListener('click', () => {
    toast.create('This is a toast notification!');
  });
</script>
```
I think avoiding shipping another API surface would be pretty nice. We could consider doing this with our toast as well

digitalsadhu:
We should involve @storsletten in the implementation early and I would love to see us collaborate with apps as well to see how they plan to implement and where we can align

wkillerud:
I had a chat with Martin about the general direction, got the goahead to make the container the live region instead of `role="status"` directly on the messages. Also goahead to remove `role="alert"` from the toast/snackbar.

We also talked about keyboard accessibility, and we'd like to explore an option of highlighting a keyboard shortcut to interact with the message, and _not_ do the tab order trickery that Web Awesome does.

digitalsadhu (commented):
oops, forgot to submit the review

packages/snackbar/docs/usage.md
wkillerud: It's Swenglish for "fast car"

packages/snackbar/docs/usage.md
wkillerud: Putting stuff on `window` is an option, if a bit icky. Might be a bit tricky to provide good types for it as well, though I think I have an example of that around somewhere.

I suspect we'll get some _constructive feedback_ if we tell people to `document.querySelector('w-snackbar')` is the only option to call `create`, but we can try.

packages/button/button.ts
```diff
@@ -305,6 +305,12 @@ class WarpButton extends FormControlMixin(LitElement) {
   }
 }
 
+declare global {
+  interface HTMLElementTagNameMap {
+    'w-button': WarpButton;
+  }
+}
+
```
wkillerud, line 314: We should do this across all components to make the TS API better when doing `createElement` and `querySelector` for Warp components.

packages/snackbar/snackbar.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    static styles = [reset, styles];
+
+    constructor() {
+        super();
+        /** Set on internals to avoid hydration mismatches in React */
+        this.internals = this.attachInternals();
+        this.internals.ariaLive = "polite";
+        this.internals.ariaRelevant = "additions";
+        this.internals.role = "log";
```
wkillerud, line 140: This doesn't seem to work for the Safari + VoiceOver combination on macOS Tahoe. macOS and Safari 26.5.

Since this component should be added to the document once (and probably by a shared HTML template) perhaps it's fine to expect users to set these attributes in light DOM? What do you think @digitalsadhu ?

wkillerud: Scratch that, adding the attributes in light DOM doesn't help. It seems Safari ignores the `w-snackbar-item` for some reason. Maybe it needs a role on internals as well.

showing the Accessibility section of the Nodes tab in Safari, showing Ignored Yes (default) for w-snackbar-item

wkillerud: I'm probably just reading it wrong, child nodes such as the text are not listed as Ignored.

packages/snackbar-item/snackbar-item.ts
```diff
@@ -0,0 +1,172 @@
...
+ * 
+ * @cssprop --w-c-snackbar-item-bg
+ * @cssprop --w-c-snackbar-item-action-bg-active
+ * @cssprop --w-c-snackbar-item-action-bg-hover
+ * 
+ * @cssprop --w-c-snackbar-item-color
+ * @cssprop --w-c-snackbar-item-action-color-active
+ * @cssprop --w-c-snackbar-item-action-color-hover
+ * 
+ * @cssprop --w-c-snackbar-item-box-shadow
```
wkillerud, line 47: Would be nice to move the source of truth for the Styling API in here, but that will have to be a TODO.

By having it here it becomes part of the custom element manifest, hover documentation in editors, and perhaps some IntelliSense smartness depending on how well the language server makes use of the information.

PR #668, opened 2026-05-13
fix(datepicker): use placeholder on safari instead of current date

### LINKED ISSUE #617: [DatePicker]: placeholder displays current date on Safari
## Descripion 
On Safari, the DatePicker component fails to display custom placeholder text (e.g., "dd/mm/yyyy"). Instead, Safari displays the current date in a muted style based on the user's system locale. This creates a confusing UX where users may think a date is already selected when the field is actually empty/undefined.

DatePicker with placeholder showing current date

## Steps to Reproduce
- Open Desktop Safari and go to [Datepicker Docs in Warp Elements Storybook](https://warp-ds.github.io/elements/?path=/docs/forms-datepicker--docs).
- Observe the empty state of the input.
- Compare with the same component in Chrome or Firefox

digitalsadhu:
On safari we use an input with type=text instead of date and add a placeholder


digitalsadhu:
Fixes https://github.com/warp-ds/elements/issues/617

packages/datepicker/datepicker.ts
```diff
@@ -105,6 +103,14 @@ class WarpDatepicker extends FormControlMixin(LitElement) {
   @property({ reflect: true })
   name: string;
 
+  /**
+   * Set a text that is shown in the date field when it doesn't have a value.
+   *
+   * Placeholder text should not be used as a substitute for labeling the element with a visible label.
+   */
+  @property({ reflect: true })
+  placeholder: string;
```
wkillerud: The way I read #617 it's about forcing cross-browser consistency in what we show when there is no value and no placeholder. @BalbinaK can correct me if I'm wrong, but I would think the need here is for `w-datepicker` to set its own default `placeholder` on the `input` that matches Chrome, so `dd.mm.åååå` for Norwegian and I don't know what for all the other locales.

BalbinaK: That's correct - we are not passing any placeholder value to the DatePicker component, expecting it to set the default the way it does in Chrome, i.e. at least `dd/mm/yyyy` and not `18/05/2026`.

Our designs did actually suggest we use a "Choose date" placeholder in the date picker, but that caused Typescript errors, so we dropped it. Perhaps allowing us to use a placeholder would suffice for now?

Btw, when I tested this branch locally in Safari using `pnpm storybook:react`, the placeholder was empty, even if I set it in the example at `datepicker.react.stories.ts` (same with `pnpm dev` and `datepicker.stories.ts`). Also when I selected a date, the value was formatted as 2026-05-18 instead of 18/05/2026 like before (whether or not it's correct - hard for me to say).

digitalsadhu:
Ok, now the default placeholder on Safari matches Chrome etc. I don't love the approach though since we are forced to not use the date input type on safari and use a type text instead... but at least it works. 

The only alternative I can think of is to try to hide the placeholder with CSS on Safari. This would avoid the confusion and since it's a date "picker"... do users really need to see the default format before selecting?

Thoughts? @BalbinaK @wkillerud

wkillerud:
> I don't love the approach though since we are forced to not use the date input type on safari and use a type text instead.

We do that already for Safari on mobile (with no placeholder I see), I reckon to opt out of the native date picker UI.

I guess either way the component is inconsistent. Either it's input type date with visual differences between browsers, or a semantic difference with differences in the type of input for visual consistency.

Since we already have special treatment for Safari I'm fine with shipping this for visual consistency, especially since it's been raised as an issue.

wkillerud (approved):

PR #669, opened 2026-05-13
docs: add missing styling.md files to all components (no content)

wkillerud:
Once we have a `.next` with this we can merge https://github.com/warp-ds/docs/pull/211 and get this out on the docsite as well.


packages/checkbox/docs/accessibility.md
wkillerud, line 1: The parent component has the docs, so I deleted these to avoid confusion/editing the wrong file in the future.

packages/checkbox/checkbox.ts
```diff
@@ -10,7 +10,7 @@ import { styles } from './styles';
 /**
  * @parent w-checkbox-group
  */
-export class WCheckbox extends FormControlMixin(LitElement) {
+export class WarpCheckbox extends FormControlMixin(LitElement) {
```
wkillerud, line 13: Components don't get picked up in the docs build unless they start with `Warp`

digitalsadhu (approved):

PR #675, opened 2026-05-20
fix(docs): add built documentation index

digitalsadhu:


package.json
```diff
@@ -140,6 +140,9 @@
       "types": "./dist/packages/*/react.d.ts",
       "import": "./dist/packages/*/react.js"
     },
+    "./docs": {
+      "default": "./dist/docs/index.md"
+    },
```
wkillerud: Is this a thing? I don't think this is a thing 😅 (importing Markdown in JS that is).

Nice to document the pathname though, but I think agents and programmatic users need to go the `path.join` and `fs.readFile` or whatever route.

digitalsadhu: Definitely not in js no. Seemed like a good idea though... or... cant hurt?

wkillerud: [Tests aren't happy about it](https://github.com/warp-ds/elements/actions/runs/26152544657/job/76923414034?pr=675#step:10:524). I think I don't want us to "cross the streams" here. Maybe there's a conventional key we can add to `package.json`? Repalce `"default"` with `"docs"` or some such?

digitalsadhu: yeah, tests being unhappy is a problem. Ill just remove and document it... document the docs

digitalsadhu: we had /docs/* from before as well... I guess remove that too

PR #676, opened 2026-05-21
chore(deps): upgrade to TypeScript 6

### LINKED ISSUE #631: Upgrade to TypeScript 6
Once at 6, consider trying a beta of 7 to see if we can gain the improved performance ahead of time.

wkillerud:
Fixes #631


tests/module-resolution/tsconfig.bundler.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "ignoreDeprecations": "6.0",
```
wkillerud, line 3: This won't fly with TS 7, so we might have to drop these tests at some point in the future.

digitalsadhu (approved):

PR #677, opened 2026-05-21
fix(icon): narrow type for icon size

josteinkringlen:
Improves the type hints for the `size` prop by adding a pixel value type and dropping the `string` type, which made the type collaps down to just `string`

### DIFF
--- a/packages/icon/icon.ts
+++ b/packages/icon/icon.ts
@@ -6,6 +6,8 @@ import { styles } from './style.js';
 // Generic parser for fetch responses
 type ResponseParser<T> = (response: Response) => Promise<T>;
 
+type PixelValue = `${number}px`;
+
 interface CacheFetchOptions<T> {
   responseParser?: ResponseParser<T>;
 }
@@ -41,7 +43,7 @@ export class WarpIcon extends LitElement {
    * @default "medium"
    */
   @property({ type: String, reflect: true })
-  size: 'small' | 'medium' | 'large' | string;
+  size: 'small' | 'medium' | 'large' | PixelValue;
 
   /** 
    * Locale used for `<title>` text.


packages/icon/icon.ts
```diff
@@ -6,6 +6,8 @@ import { styles } from './style.js';
 // Generic parser for fetch responses
 type ResponseParser<T> = (response: Response) => Promise<T>;
 
+type PixelValue = `${number}px`;
```
wkillerud, line 9: That's cool, didn't know you could do this in types 🙌

packages/icon/icon.ts
```diff
@@ -41,7 +43,7 @@ export class WarpIcon extends LitElement {
    * @default "medium"
    */
   @property({ type: String, reflect: true })
-  size: 'small' | 'medium' | 'large' | string;
+  size: 'small' | 'medium' | 'large' | PixelValue | (string | {});
```
wkillerud: I'm having trouble getting code completion unless we remove ` | (string | {})` 🤔

Since it's "just" types I think we're fine making them stricter without doing a semver major/breaking change. I did a quick search of our internal usage and it's mostly the `small | medium | large` keywords and a few px values here and there. I saw ~5 uses of just a number, which I don't mind discouraging.

Which is to say I think we can remove ` | (string | {})` to get the improved code completion and stricter types.

wkillerud:
Nevermind the storybook thing, it's because the PR is from a fork

packages/icon/icon.ts
josteinkringlen: Ah lol, had a typo 🤦🏻‍♂️ It's supposed to be `(string & {})` not `(string | {})` 😄 Pushed a change for it now which should fix the code completion 🙂 But I can remove it entirely instead 😄 

Edit: `(string & {})` has been removed

wkillerud (approved):

wkillerud:
Thank you! This'll be shipped in [2.10.0-next.8](https://github.com/warp-ds/elements/releases) right away. 2.10.0 stable will probably go out some time next week.

PR #682, opened 2026-05-22
fix: compile using browserslist when bundling

wkillerud:
Comparing before and after I got `static{` in the bundle before adding browserslist, and no `static{` (plenty of `.static{` CSS though) in `eik/index.js`.


wkillerud:
I see we have a bunch of other small build scripts that use `esbuild` but target `es2018`. I can add [browserslist-to-esbuild](https://www.npmjs.com/package/browserslist-to-esbuild) to wire them up as well.

digitalsadhu (approved):

PR #687, opened 2026-05-26
feat(types): add improved React type support to types

digitalsadhu:
Discovered that properly typing the custom elements usage in recommerce-bet-front resulted in quite a bit of type wrapping so this PR is me getting Codex to patch the types output to get a "it just works" result in React apps. Ive tested the build output in recommerce-bet-front and it seems to be working. 

The issues were:
1. using react properties with Warp elements were not supported out of the box. eg.

```jsx
<w-button onClick={...} onBlur={...} className="..." style={...} />
<w-breadcrumbs suppressHydrationWarning dangerouslySetInnerHTML={...} />
<w-attention ref={...} popover />
```

2. name clash for popover
Once we added in the React types, popover is already in there and so our popover in Warp Elements gets overwritten.
The fix for this is nasty 
```ts
type ReactCustomElementProps<T extends HTMLElement, Props> =
  Omit<BaseProps<T> & BaseEvents, keyof ReactElementProps<T> | keyof Props> &
  Omit<ReactElementProps<T>, keyof Props> &
  Props;
```
...essentially, remove all props and re-ad them with the Warp props last so they win.

So this PR makes it so Warp Elements owns the React JSX custom-element typings, letting consuming apps delete their hand-written React augmentation shims and rely on the package as the single source of truth.

The code here in my opinion is pretty much unreadable garbage but I only partly blame AI for this. The biggest culprit in my mind is the combo of the sheer complexity of this kind of typing combined with the procedural generation of these types out of JS...

I feel dirty.

### DIFF
--- a/.cem/custom-element-manifest.config.js
+++ b/.cem/custom-element-manifest.config.js
@@ -31,6 +31,8 @@ function reactJsxIntrinsicPropsPlugin() {
         throw new Error('Unable to derive custom element types from dist/index.d.ts');
       }
 
+      // Keep the generated CustomElements/Solid types intact and add a React-only
+      // intrinsic map where React props and component props are merged by precedence.
       const reactCustomElementsTypes = `type ReactElementProps<T extends HTMLElement> = import("react").DOMAttributes<T> &
   import("react").AriaAttributes &
   import("react").RefAttributes<T> &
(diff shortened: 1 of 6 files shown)


wkillerud (approved):
I account swapped on GitHub.com to approve this message

digitalsadhu:
> I account swapped on GitHub.com to approve this message

but would you say that this looks good?

wkillerud:
bugs bunny saying no

digitalsadhu:
k-fine-disappointed-sad-monkey-pissed-angry

PR #689, opened 2026-05-29
fix: regenerate styles.ts with unsafeCSS after PR #685

turkus:
## Summary

PR #685 updated `build/styles.js` to use `unsafeCSS()` instead of `css\`\`` tagged template literals, but the generated `styles.ts` files in each component package were not regenerated.

This caused SWC's backslash escaping corruption to persist — `replaceSync()` silently drops all CSS rules after the first corrupted arbitrary-value selector (e.g. `\\[` reduced to `[`), making components like `w-switch` invisible because their utility CSS (`w-44`, `h-24`, `block`, etc.) never reaches the shadow DOM's `adoptedStyleSheets`.

**Fix:** ran `pnpm build:styles` to regenerate all `styles.ts` files with the new `unsafeCSS()` format.

## Verification

- Confirmed `w-switch` shadow DOM stylesheet has only 2 CSS rules on pre-prod (should have ~50+)
- After regeneration, all 17 component `styles.ts` files now use `unsafeCSS()` instead of `css\`\``

## Suggestion

Consider adding `pnpm build:styles` to the main `build` script in `package.json` so generated styles are always up-to-date when publishing.


wkillerud (approved):

wkillerud:
Thank you!

> Consider adding pnpm build:styles to the main build script in package.json so generated styles are always up-to-date when publishing.

We're gradually replacing our internal use of UnoCSS with our own stylesheets so the components are more easily customizable (f. ex. [for radio group](https://github.com/warp-ds/elements/blob/next/packages/radio-group/radio-group-styles.ts), [docs](https://github.com/warp-ds/elements/blob/next/packages/radio-group/docs/styling.md)), after that this won't be a problem anymore.

digitalsadhu (commented):
Looks good, only thing I want to be slightly careful about is that we are using the style builds as a temporary measure while we replace them with styles written for each component (rather than generated from Uno) and I want to just go through the diffs here and make sure nothing weird is sneaking in anywhere before I approve.

trygve-lie:
Sorry for being a bit late to the party here. I am not super comfy with using `unsafeCSS()` to be honest. There are security concerns in this aspect.

Are there other ways to solve this?

wkillerud:
We don't pass outside input to `unsafeCSS` in this case, I think it's fine. As mentioned we're working on replacing this whole build step.

PR #693, opened 2026-06-04
refactor(inputs): implement shared token styling

Skadefryd:
Consolidates shared input styling tokens for textfield, textarea, and select
Replaces utility-class-based select styles with element selector rules
Integrates select help text into the component’s built-in styling API instead of a separate shadow-DOM helper.
+ css api docs


wkillerud (approved):
Looks good to me comparing the Storybook preview to what's on `next` 👍 

No CSS parts API for these components?

packages/select/input-styles.ts
```diff
@@ -0,0 +1,76 @@
...
+    --_color: var(--w-c-input-label-color, var(--w-s-color-text));
+    --_font-size: var(--w-c-input-label-font-size, var(--w-font-size-s));
+    --_line-height: var(--w-c-input-label-line-height, var(--w-line-height-s));
+    --_font-weight: var(--w-c-input-label-font-weight, 700);
+    --_padding-bottom: var(--w-c-input-label-padding-bottom, 0.4rem);
+    --_cursor: var(--w-c-input-label-cursor, pointer);
+    --_display: var(--w-c-input-label-display, block);
+
+    /* Apply styles */
+    display: var(--_display);
```
digitalsadhu: why do we create these internal tokens eg. --_color and not just use the public api tokens directly?

digitalsadhu: I guess just because we use them multiple times?

packages/textarea/textarea.ts
```diff
@@ -269,7 +263,7 @@ class WarpTextarea extends FormControlMixin(LitElement) {
     this.#clearValidationState();
   }
 
-  static styles = [reset, styles];
+  static styles = [reset, styles, inputLabelStyles, inputHelpTextStyles];
```
digitalsadhu: do we still need a generic reset?

packages/textfield/textfield.ts
```diff
@@ -254,7 +266,7 @@ class WarpTextField extends FormControlMixin(LitElement) {
   // ::slotted([Simple Selector]) confirms to Specificity rules, but (being simple) does not add weight to lightDOM skin selectors,
   // so never gets higher Specificity. Thus in order to overwrite style linked within shadowDOM, we need to use !important.
   // https://stackoverflow.com/a/61631668
-  static styles = [reset, styles, wTextfieldStyles];
+  static styles = [reset, styles, wTextfieldStyles, inputLabelStyles, inputHelpTextStyles];
```
digitalsadhu: loving the reusable modular approach

digitalsadhu (approved):

packages/select/input-styles.ts
Skadefryd: The --_internal token pattern provides a few benefits:

1. Centralized token resolution - All public API tokens and their fallbacks are declared in one place (:host), making the styling API easy to see at a glance, while the actual CSS rules stay clean: color: var(--_color) instead of repeating color: var(--w-c-label-color, var(--w-s-color-text)) everywhere.

2. State-based overrides without duplication - You can override internal tokens for specific states without repeating the full fallback chain:


:host([invalid]) {
  --_color: var(--w-c-help-text-color-invalid, var(--w-s-color-text-negative));
}
If you used public tokens directly in rules, you'd have to repeat var(--w-c-help-text-color, var(--w-s-color-text-subtle)) everywhere and couldn't easily override for states.

3. Reusability - If multiple elements need the same value, you reference the internal token once instead of duplicating the public token + fallback chain across multiple rules.

I understand that you ask since these benefits are not always apparent when looking at a component in isolation.  These benefits varies a bit from component to component, but I think its a good idea to use the same pattern in all of them.  This pattern maxes out its benefits in components that has a lot of variants, like a button, or a badge. Should lead to way less selector repetition as a whole , and rather manipulate tokens instead.

packages/textarea/textarea.ts
Skadefryd: The reset is a very good question, I was wondering the same thing.  I would personally not drop the reset if I wasnt sure it was excisting through other means. Becasue... removing it introduces unpredictability to the equation. Even if each component is tested and cleared as 'working as expected without a reset' (Thats the whole supported browserstack test), you would have to do this on every single change for all time cause anything could potentially trigger something that is weird in some browser.  VS knowing that the reset just fixes browser differences for you and not having to think of it.

Skadefryd: its a cost/benefit analysis thing i suppose. Some components might be easy to see if they will need it or not , others not.

packages/select/input-styles.ts
digitalsadhu: Makes sense

PR #696, opened 2026-06-10
fix: add a w- prefix to the data attribute

wkillerud:
To avoid naming conflicts down the line, consistency.


digitalsadhu (approved):
I guess we might want to take a backwards compatible round and add w prefixes to any other data attribute usages as well

Issue #357, opened 2025-10-22
SEO audit


digitalsadhu:
The SEO team raised a concern around client side rendering and SEO. 
If links are hidden inside a component then AI crawlers cannot see them as they do not render client side JS.

Example

Bad
```html
<!-- Provides some sort of links to other things-->
<w-thing></w-thing>
```

Good
```html
<!-- Link components are provided to the component and therefore stay in the initial server render-->
<w-thing>
  <a href="/one">link one</a>
  <a href="/two">link two</a>
</w-thing>
```

Many components will not be affected but components such as w-pagination do render links internally and are therefore problematic.

Worth noting that, we only care for public facing pages. Also, there will be an impact for developers if they are forced to write more verbose components for SEO.

Steps required:
- Audit components for SEO compatibility 
- Decide on a solution for incompatible components
  - we could provide an alternative syntax for components that need SEO support added
  - OR we could try to re-engineer incompatible components

PR #698, opened 2026-06-12
fix(breadcrumbs): avoid hydration mismatch in breadcrumbs by adding a w-breadcrumb-item component

digitalsadhu:


packages/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -0,0 +1,45 @@
+import { LitElement, html, nothing, css } from 'lit';
+import { property } from 'lit/decorators.js';
+
+class WarpBreadcrumbItem extends LitElement {
```
wkillerud, line 15: The class could use some JSDoc (especially `@parent w-breadcrumbs` so the docs build is happy).

packages/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -0,0 +1,45 @@
...
+      color: var(--w-s-color-text)
+    }
+    .s-text-link{
+      color: var(--w-s-color-text-link);
+      text-decoration: none;
+    }
+    .s-text-link:hover {
+      text-decoration: underline;
+    }
+    .separator{
```
wkillerud: Should perhaps use the new CSS pattern for new components?

packages/breadcrumbs/docs/accessibility.md
```diff
@@ -8,33 +8,23 @@ The component has a localized default accessible label. Use `aria-label` when th
...
-```
-
-If the current page remains a link, keep `aria-current="page"` on that link:
-
-```html
-<w-breadcrumbs aria-label="You are here">
-  <a href="/real-estate">Real estate</a>
-  <a href="/real-estate/homes">Homes for sale</a>
-  <a href="/real-estate/homes/oslo" aria-current="page">Oslo</a>
+  <w-breadcrumb-item href="/real-estate">Real estate</w-breadcrumb-item>
```
wkillerud, line 25: Maybe fine for breadcrumbs, but I saw #357 in the backlog. This might be a regression on that front, moving `<a>` out of Light DOM.

packages/breadcrumb-item/breadcrumb-item.ts
digitalsadhu: Was thinking about it but figured we'd take all of breadcrumbs in a later pass. But you're probably right, I could at least do breadcrumbs-item now

wkillerud (approved):

PR #700, opened 2026-06-15
docs(radio): add styling API documentation and remove unused classes

Skadefryd:
Added comprehensive styling documentation for radio component that was previously marked as partial (had tokens but missing docs).

Changes:
- Created docs/styling.md with all component tokens documented:
  - Label tokens (font-size, line-height, color, color-disabled)
  - Control tokens (size, gap, radius, border-width, checked-border-width)
  - Color tokens for all states (default, checked, invalid, disabled)
  - Focus tokens (outline width, color, offset)
  - Cursor tokens (default, disabled)
  - Animation tokens (transition)
- Documented all parts: control, label
- Added usage examples for common customizations
- Explained rich label content via slot (same as checkbox)
- Noted high contrast mode support
- Explained custom implementation using ElementInternals vs native input

Code cleanup:
- Removed unused class attributes (control, label)
- Kept wrapper class (used by CSS for layout)
- No functional changes - parts remain exposed for customization

Architecture documentation:
- Added implementation note explaining custom role="radio" approach
- Added TODO noting checkbox inconsistency (uses hidden native input)
- Suggests future alignment after backwards compatibility assessment

Completes Phase 2 radio documentation.


packages/radio/radio-styles.ts
```diff
@@ -50,12 +50,6 @@ export const styles = css`
     --w-c-radio-cursor: var(--w-c-radio-cursor-disabled, not-allowed);
   }
 
-  .wrapper {
-    display: inline-flex;
-    align-items: center;
-    gap: var(--_gap);
-  }
-
```
wkillerud: This doesn't get replaced by a `[part="base"]`, that seems like a potentially breaking change?

digitalsadhu (approved):

packages/radio/radio-styles.ts
Skadefryd: I think these were moved to :host and inherited from there instead. 
`
:host {
  display: inline-flex;
  align-items: center;
  gap: var(--_gap);
`

wkillerud: Aha, I didn't expand the diff far enough up 🙈

wkillerud (approved):

packages/radio/radio-styles.ts
Skadefryd: Hmm... I think you might be right, the flex stuff might be problematic, it seams to be set on the wrong level now.  I think it was set onmultiple levels before. Retesting this stuff, and prob pushing a fix

PR #702, opened 2026-06-15
fix(tabs): make it easier to do React hydration correctly

wkillerud:
I think this is the best we can do for #594 , guide people to set `id` and `active` props themselves.

The slot should always be set to these values I suggest we add in the wrappers, but if they really want to users can override them.


tests/react-hydration.ts
```diff
@@ -54,10 +54,7 @@ function nodeToReact(node: Node): React.ReactNode {
 		const el = node as Element;
 		const props: Record<string, unknown> = {};
 		for (const attr of el.attributes) {
-			// Convert HTML attributes to React props
-			let name = attr.name;
-			if (name === "class") name = "className";
-			if (name === "for") name = "htmlFor";
```
wkillerud, line 60: I ended up spinning up a test application to do SSR + hydration and this was the culprit for the failing hydration test.

The diff in the error message was super confusing, it essentially said `htmlFor="tab1"` was replaced with `htmlFor="tab1"`. [They're the same.](https://media0.giphy.com/media/v1.Y2lkPTc5MGI3NjExajZqeTVmaGFseTUzdTZpM2F3YjZob2ZtZGNtMXlrdW5rMnV6ZDR6dCZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/pV0lVLeA0JXjBiO5Cp/giphy.gif)

When I tried to render `<w-tab htmlFor="tab1">` in React `htmlFor` was not mapped to `for` in HTML. I didn't find a good source for why. Either it's a custom elements thing, or React stopped calling it `htmlFor` in 19 🤷

wkillerud: [htmlFor](https://react.dev/reference/react-dom/components/common#:~:text=on%20virtual%20keyboards.-,htmlFor%3A%20A%20string.,-For%20%3Clabel%3E) is still a thing, so something else weird is up here.

PR #708, opened 2026-06-18
fix(breadcrumb-item): add missing default css values

digitalsadhu:


joms (approved):

wkillerud (approved):
LGTM once the linter is happy.

For reference the problem was that without a fallback for the component tokens the [var() resolves to `unset`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/var#invalid_values) if the component token is not defined.

PR #717, opened 2026-06-22
docs: add controls for more props in React land

wkillerud:
Testing Optional prop, seems to work. If it's both required and optional, required wins.


digitalsadhu (commented):
Nice

PR #718, opened 2026-06-23
fix: narrow the types for event handlers in React

wkillerud:
Also add all the custom elements to the global HTML tag list.


packages/textfield/react.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+test("onChange and onInput have the expected types", async () => {
+	const onChange = vi.fn();
+	const onInput = vi.fn();
+
+	const page = await render(
+		<TextField
+			name="test"
+			label="React TextField"
+			onChange={(e) => {
+				onChange(e.target.value);
```
wkillerud, line 17: Getting this `onChange` without doing type casting was the goal.

packages/textfield/textfield.ts
```diff
@@ -349,13 +346,37 @@ class WarpTextField extends FormControlMixin(LitElement) {
...
+			value,
+			target: e.target,
+		};
+		const deprecationProxy = new Proxy(detail, {
+			get(target, prop) {
+				if (typeof window !== "undefined") {
+					// be quiet in prod
+					if (!window.location.host.startsWith("www")) {
+						console.warn(
+							"w-textfield's CustomEvent is deprecated, please use the browser-native events instead (e.g. replace e.detail.value with e.target.value)",
```
wkillerud, line 369: Snuck in a cheeky little deprecation warning here to nudge folks over (including ourselves in combobox).

packages/select/select.ts
```diff
@@ -396,17 +398,21 @@ export class WarpSelect extends FormControlMixin(LitElement) {
...
 
 	// Fire a custom 'change' event, used when the dropdown changes state.
-	/**  @internal */
-	onChange(event: Event) {
+	private onChange(event: Event) {
 		const target = event.currentTarget as HTMLSelectElement;
 		const nextValue = target.value;
 
-		// Avoid forwarding the internal native change event and emitting two host events.
-		event.stopPropagation();
```
wkillerud, line 405: This does lead to two host events, but I would rather have that than no native events.

We have double events in `w-textfield` already.

digitalsadhu (approved):
LGTM. I basically skimmed and I defer to your superior TS knowledge

packages/select/select.ts
wkillerud: Couldn't add the same deprecation warning because we set `event.detail` to be the raw string `value`.

packages/modal-header/modal-header.ts
```diff
@@ -158,6 +158,12 @@ export class WarpModalHeader extends CanCloseMixin(LitElement) {
 /** @deprecated Exported for backwards compatibility. Use WarpModalHeader. */
 export const ModalHeader = WarpModalHeader;
 
+declare global {
+	interface HTMLElementTagNameMap {
+		"w-modal-header": WarpModalHeader;
+	}
+}
+
```
digitalsadhu, line 166: It does my head in that we keep having to add these various things and it never feels like there is one definitive way to do this.

package.json
```diff
@@ -32,117 +32,9 @@
...
-    "./react/step": {
-      "types": "./dist/packages/step/react.d.ts",
-      "import": "./dist/packages/step/react.js"
-    },
-    "./react/step-indicator": {
-      "types": "./dist/packages/step-indicator/react.d.ts",
-      "import": "./dist/packages/step-indicator/react.js"
+     "./react": {
+      "types": "./dist/react.d.ts",
+      "import": "./dist/react.js"
```
digitalsadhu, line 37: fair enough, cant think of any issues with this other than maybe bundling in more wrappers than are used to each apps bundle? I guess hopefully tree shaking will take care of it

wkillerud: Wrappers are pretty small, I think we should be fine.

PR #719, opened 2026-06-23
feat: add styling API to card

wkillerud:
Fixes WARP-1389

Designers love me

a purple colored card with cyan text and magenta border, it looks fantastic


digitalsadhu (approved):
Nice!

PR #723, opened 2026-06-26
feat(modal): add --w-c-modal-padding-bottom, ModalContent React wrapper

wkillerud:
- Apply the default slot value in the Modal React wrapper components
- Add a `ModalContent` for convenience
- Add `--w-c-modal-padding-bottom` so users can get a scroll that goes to the screen edge on mobile when there's no footer (favorites modal, the "footer" is a pagination component that scrolls into view).

For WARP-1397

### DIFF
--- a/packages/modal/react.ts
+++ b/packages/modal/react.ts
@@ -50,11 +50,11 @@ type ModalProps = Omit<
  *     </Button>
  *     <Modal show={open} id="example-modal-one" onHidden={() => setOpen(false)} onShown={() => setOpen(true)}>
  *       <ModalHeader slot="header" title="An example modal" />
- *       <div slot="content">
+ *       <ModalContent>
  *         <p>
  *           <!-- modal content -->
  *         </p>
- *       </div>
+ *       </ModalContent>
  *       <ModalFooter slot="footer">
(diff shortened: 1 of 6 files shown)


packages/modal/react.ts
```diff
@@ -50,11 +50,11 @@ type ModalProps = Omit<
  *     </Button>
  *     <Modal show={open} id="example-modal-one" onHidden={() => setOpen(false)} onShown={() => setOpen(true)}>
  *       <ModalHeader slot="header" title="An example modal" />
```
wkillerud:
```suggestion
 *       <ModalHeader title="An example modal" />
```

packages/modal/react.ts
```diff
@@ -50,11 +50,11 @@ type ModalProps = Omit<
...
- *       <div slot="content">
+ *       <ModalContent>
  *         <p>
  *           <!-- modal content -->
  *         </p>
- *       </div>
+ *       </ModalContent>
  *       <ModalFooter slot="footer">
```
wkillerud:
```suggestion
 *       <ModalFooter>
```

wkillerud (commented):
Can remove the slots that are set by default now from the example

PR #726, opened 2026-07-02
fix(slider): don't show suffix if tooltip value is Min or Max

wkillerud:
Fixes a bug found in design review of the usage guideline docs


wkillerud:
Little bit of "there I fixed it", but it would be a breaking change to all of a sudden require users provide the suffix themselves.

digitalsadhu (approved):

PR #727, opened 2026-07-07
feat(tooltip): create a new tooltip component 

wkillerud:
For WARP-1382

To verify it works for what we need it to internally I added a `tooltip` prop to `w-textfield`. Thinking I'll do the pass to add it to all form components before merging once we're happy with the tooltip component itself.


packages/tooltip/tooltip.test.ts
```diff
@@ -0,0 +1,15 @@
+import { html } from "lit";
+
+import { expect, test } from "vitest";
+import { render } from "vitest-browser-lit";
+
+import "./tooltip.js";
+
+test("does the thing", async () => {
```
wkillerud: It does indeed do the thing. Need some tests here! 🤦

packages/tooltip/tooltip.hydration.test.ts
```diff
@@ -0,0 +1,19 @@
+import { afterEach, beforeEach, describe, expect, test } from "vitest";
+import {
+	setupHydrationWarningCapture,
+	testHydration,
+} from "../../tests/react-hydration.js";
+
+import "./tooltip.js";
+
+describe("w-snackbar React SSR hydration", () => {
```
wkillerud:
```suggestion
describe("w-tooltip React SSR hydration", () => {
```

packages/textfield/textfield.ts
```diff
@@ -213,6 +213,16 @@ class WarpTextField extends FormControlMixin(LitElement) {
...
 
+	/**
+	 * Suplementary information that should show in a tooltip behind an information icon after the label.
+	 *
+	 * Use the `tooltip` slot if you need markup and not just text.
+	 *
+	 * You must provide a label to be able to show an info icon with a tooltip.
+	 */
+	@property({ type: String, reflect: true })
+	tooltip?: string;
```
wkillerud, line 224: Could use a test for this new property in textfield

packages/tabs/tabs.react.stories.tsx
```diff
@@ -28,14 +28,14 @@ export default {
...
+	<Tabs active="panel1">
+		<Tab id="tab1" htmlFor="panel1">
 			First Tab
 		</Tab>
-		<Tab id="tab2" for="panel2">
+		<Tab id="tab2" htmlFor="panel2">
 			Second Tab
 		</Tab>
-		<Tab id="tab3" for="panel3">
+		<Tab id="tab3" htmlFor="panel3">
```
digitalsadhu, line 38: whats going on here with the change from for to htmlFor?

packages/tooltip/docs/styling.md
```diff
@@ -0,0 +1 @@
+## Styling
```
digitalsadhu: to be added?

packages/tooltip/tooltip.ts
```diff
@@ -0,0 +1,521 @@
...
+		if (!this.noArrow) {
+			description += " ";
+			switch (this.#actualPlacement) {
+				// The arrow points in the opposite direction from the placement
+				case "top-start":
+				case "top":
+				case "top-end":
+					description += i18n._({
+						id: "tooltip.aria.pointingDown",
+						message: "pointing down",
```
digitalsadhu, line 129: should this be "arrow pointing down"? or is "pointing down" ok?

digitalsadhu (approved):
This looks really great. Love the API surface and looks purrrrdy in storybook.

packages/tabs/tabs.react.stories.tsx
wkillerud: I'm a bit surprised `for` even worked to be honest. [htmlFor is what you typically use in React land](https://react.dev/reference/react-dom/components/common#:~:text=React%20uses%20the%20standard%20DOM%20property%20names).

packages/tooltip/tooltip.ts
wkillerud: The sentence starts with "Speech bubble" so I think this is right. It's what we had in the Attention component.

packages/textfield/textfield.ts
wkillerud: I'll do the rest of the forms stuff in a separate PR

PR #729, opened 2026-07-09
feat: add info icon with tooltip to form fields

wkillerud:
Used with the `tooltip` property on all form fields.

Datepicker was missing a bunch of stuff compared to the other form fields, so added that as well.


packages/checkbox-group/checkbox-group.ts
```diff
@@ -142,6 +150,20 @@ export class WarpCheckboxGroup extends FormControlMixin(LitElement) {
...
+			background: transparent;
+			border: none;
+			height: 16px;
+			margin: 0 0 0 4px;
+			padding: 0;
+			vertical-align: text-top;
+		}
+
+		w-tooltip {
+			display: inline-block;
```
wkillerud, line 165: Without this `inline-block` we get some weird additional height compared to labels without the tooltip.

digitalsadhu (approved):
Looks good!

PR #731, opened 2026-07-10
feat: min/maxlength, JSX helpText, bugfix for onchange in textarea

### LINKED ISSUE #720: [Textarea] add maxLength/minLength property support
## Description

The native `<textarea>` element supports a `maxLength` attribute that prevents user input beyond a specified character limit. The `WarpTextarea` web component doesn't expose this property, so it's not available in the React wrapper's type definitions either. There might be more attributes 

## Current behavior

Setting `maxLength` on `<Textarea>` produces a TypeScript error:
  ```ts
Property 'maxLength' does not exist on type '...'
```

## Expected behavior

`WarpTextarea` should declare a `maxLength` property that reflects to the inner `<textarea>` element's `maxlength` attribute, matching native behavior.

### LINKED ISSUE #721: [Textarea] Support rich content in helpText (slot or ReactNode type)
### Description

  The `@warp-ds/react` `TextArea` component accepted `ReactNode` for the `helpText` prop, allowing inline links,
  formatting, and translated JSX (e.g. `<Trans>` with embedded `<a>` tags). The `@warp-ds/elements` `Textarea` only
  accepts `string | undefined`.

  ### Use case

  We have a feedback form where the help text includes a character counter and a link to customer service:

  ```tsx
  // Previously worked with @warp-ds/react TextArea
  <TextArea
    helpText={
      <Trans
        id="itemPageLayout.feedback.helpText"
        message="{charCount} characters left. If you need help, please contact our <link>customer service</link>"
        values={{ charCount: maxChars - charCount }}
        components={{
          link: <a href={helpCenter} target="_blank" rel="noopener noreferrer" />,
        }}
      />
    }
  />
  ```
Check the source code [here](https://github.com/m10s-green/aurora-mobility-private-buyer/blob/2c346b35f1f66e3ee8fcc3881b56c677901c2473/apps/item-page-layout/src/client/components/FeedbackForm.tsx#L268-L284).

Image

### LINKED ISSUE #722: [Textarea] onChange doesn't fire on React wrapper
The inner `<textarea>` fires a native change event on blur, but it has `composed: false` so it never crosses the shadow DOM boundary. The React wrapper maps `onChange` to `"change"` on the host `<w-textarea>`, so the handler never fires.

`TextField` solves this by re-dispatching the event from its handler. Could a similar approach work for `Textarea` - e.g. adding `@change` on the inner `<textarea>` and re-dispatching on the host?

In the meantime, `onInput` works as a workaround (fires on every keystroke instead of on blur).

wkillerud:
Fixes these issues:

- #720 
- #721 
- #722


BalbinaK (approved):
🙏

digitalsadhu (approved):
Found and fixed a couple issues, now dealing with merge conflicts then ill squash

PR #735, opened 2026-07-15
fix(pagination): add missing last page button, visual fixes

### LINKED ISSUE #424: Pagination
Critical:

* [ ] No mobile version.

Findings:

* [ ] Check if text is correct, it should be Body-Strong font style (Bold version).
* [ ] Check if icon is a new WARP icon library.
* [ ] Is it possible to add "End" arrows?

Image

https://nmp-jira.atlassian.net/browse/WARP-1233

https://miro.com/app/board/uXjVLmvdEfA=/?moveToWidget=3458764644357328167&cot=14

### LINKED ISSUE #468: Set host of pagination to `display: block`
By the looks of it, it seems like all consumers of `Pagination` will have to add `display: block` to the host. Therefore it should be supplied from the design system itself.

wkillerud:
Also includes some needed improvements and bugfixes related to the `page-click` event handler.

- We couldn't cancel the navigation without a generic click handler as well which was annoying.
- Clicks on non-page elements in the component fired the event, but with `null` as the value.
- The value was always a number, but passed as a string. Let's do the parsing for users.

For WARP-1386, WARP-1387


digitalsadhu (approved):
**Inspect packages/pagination/docs/examples.md, packages/pagination/docs/usage.md, packages/pagination/pagination.stories.ts, packages/pagination/pagination.test.ts, packages/pagination/pagination.ts for blast radius and ownership.**

These files all look good to me.

**Review changed tests before implementation and verify assertions still prove the requirement.**

Tests appear to prove the requirement.

PR #743, opened 2026-07-30
fix(modal-header): constrain close button column width to prevent title overflow

### LINKED ISSUE #742: [ModalHeader] Long title overflows behind close button when it doesn't wrap
## Problem

When a `ModalHeader` shows a close button, a sufficiently long title - one that extends into the close button area but isn't long enough to wrap - overflows behind the button, making the text and button overlap:

Image

## Root cause (suspected)

The close button uses `position: absolute`, so it's removed from grid flow. In the show-close (without back button) configuration, the grid template lacks explicit column sizing:

```css
  [part="header"][show-close]:not([show-back]) {
      grid-template:
          "top top top"
          "title title close";
      /* No column sizes defined — title expands to full width */
  }
```

Because no column track widths are declared, the title area fills the full container width. The absolute-positioned close button then floats on top of it with no reserved space accounted for.

By contrast, the show-back + show-close variant already handles this correctly by declaring explicit column widths:

```css
  [part="header"][show-back][show-close] {
      grid-template:
          "top top top"
          "back title close" / 44px 1fr 44px;  /* ✓ title constrained to 1fr */
  }
```

Image
(BTW, the back button seems to have inconsistent padding compared to the close button)

## Proposed fix

Apply the same column sizing pattern to the other affected variants:

```css
  [part="header"][show-close]:not([show-back]) {
      grid-template:
          "top top top"
          "title title close" / 1fr 1fr 44px; /* or: auto 1fr 44px */
  }
```

Image

BalbinaK:
Fixes #742 - long modal header titles no longer overflow behind the close button.

## Summary

Added CSS Grid column sizing (`/ 1fr 1fr 44px`) to the modal header layout, constraining the close button to a fixed 44px column so the title truncates correctly instead of overlapping it.

### Before
Screenshot 2026-07-30 at 14 29 03

### After
Screenshot 2026-07-30 at 14 28 32

### DIFF
--- a/packages/modal-header/styles.ts
+++ b/packages/modal-header/styles.ts
@@ -35,7 +35,7 @@ export const styles = css`
 	[part="header"][show-close]:not([show-back]) {
 		grid-template:
 			"top top top"
-			"title title close";
+			"title title close" / 1fr 1fr 44px;
 	}
 
 	[part="header"][show-back][show-close] {


BalbinaK:
Not sure what's the matter with the failing test - it also occurs on `next`. I'd appreciate some help here 🙏 Same in my other [PR](https://github.com/warp-ds/elements/pull/743).

digitalsadhu:
> Not sure what's the matter with the failing test - it also occurs on `next`. I'd appreciate some help here 🙏 Same in my other [PR](https://github.com/warp-ds/elements/pull/743).

Just a bit of test flakiness unfortunately, nothing you've done :)

digitalsadhu (approved):
Nice catch, thanks for the fix!

PR #744, opened 2026-07-30
fix(button): apply custom focus outline styles instead of browser default

BalbinaK:
## Summary

- Moves `:focus-visible` from `:host(:focus-visible) button[part="base"]` to `button[part="base"]:focus-visible`
- When placed on `:host`, `:focus-visible` does not propagate into shadow DOM - the browser ignored the custom `--_outline-width`, `--_outline-color`, and `--_outline-offset` variables and fell back to its default focus outline instead
- The fix ensures the design system's focus outline styles are applied correctly on keyboard navigation

### Before
Screenshot 2026-07-30 at 14 27 46Screenshot 2026-07-30 at 14 27 50

### After
Screenshot 2026-07-30 at 14 27 30
Screenshot 2026-07-30 at 14 27 36

### DIFF
--- a/packages/button/styles.ts
+++ b/packages/button/styles.ts
@@ -263,7 +263,7 @@ export const wButtonStyles = css`
 		color: var(--_color-active);
 	}
 
-	:host(:focus-visible) button[part="base"] {
+	:host button[part="base"]:focus-visible {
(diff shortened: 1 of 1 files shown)


digitalsadhu (approved):
Fantastic thanks!

PR #750, opened 2026-08-10
feat: include form submitter via new warpSubmitContext event property

digitalsadhu:
## Intent

  When Warp form components (textfield, select, checkbox, etc.) trigger implicit form submission via Enter key, the native SubmitEvent.submitter is not populated because form.requestSubmit() is called without
   a submitter argument. This breaks scenarios where server frameworks or client code rely on the submitter's name/value to determine intent (e.g., which button was "clicked").

  This PR attaches a warpSubmitContext object to SubmitEvents that provides:
  - The Warp element that initiated submission (initiator)
  - The native submitter passed to requestSubmit() if any (nativeSubmitter)
  - The form's default submit button per HTML spec (defaultSubmitter)

##  Approach

  - Added a new packages/utils.ts module with helper functions for form submission
  - requestSubmitWithDefaultSubmitter() finds the form's default submit button (first submit button in DOM order, following HTML spec) and passes it to requestSubmit()
  - Uses a WeakMap to track pending submit context, then decorates the SubmitEvent with warpSubmitContext via a capturing submit listener
  - Updated all form components (textfield, select, checkbox, radio, datepicker, slider) to use requestSubmitWithDefaultSubmitter() for implicit Enter-key submissions
  - Updated w-button to use requestSubmit() with itself as initiator, plus added null-safety for form association
  - Exported WarpSubmitContext type for consumer type safety

##  Risk

  - Form submission behavior change: Implicit submissions now include the default submitter's name/value in FormData, which matches native browser behavior but could surface previously-missing data
  - New API surface: event.warpSubmitContext is a new property consumers may depend on

##  Verification

  - [x] Tests added or updated
  - [x] Existing tests pass (1 unrelated flaky tooltip test)
  - [ ] Manual verification completed, if relevant
  - [ ] Screenshots or recordings attached, if UI changed

##  Commands run:

npm test -- packages/button/button.test.ts packages/textfield/textfield.test.ts packages/select/select.test.ts packages/checkbox/checkbox.test.ts
- 149 passed, 1 skipped

##  AI involvement

  Yes. AI assisted with designing the warpSubmitContext approach, implementing the utility functions in packages/utils.ts, updating all affected components, and writing test cases. The implementation was
  verified by reviewing diffs and running the test suite.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,10 @@
       "types": "./dist/packages/toast/api.d.ts",
       "import": "./dist/api.js"
     },
+    "./utils": {
+      "types": "./dist/packages/utils.d.ts",
+      "import": "./dist/packages/utils.js"
+    },
     "./custom-elements.json": "./dist/custom-elements.json",
     "./react": {
       "types": "./dist/entrypoint-react.d.ts",
@@ -52,6 +56,9 @@
   "types": "./dist/index.d.ts",
   "typesVersions": {
     "*": {
+      "utils": [
+        "./dist/packages/utils.d.ts"
+      ],
       "react/affix": [
         "./dist/packages/affix/react.d.ts"
       ],
--- a/packages/button/button.test.ts
+++ b/packages/button/button.test.ts
@@ -21,7 +21,40 @@ test("by default button type is button", async () => {
 		.toHaveAttribute("type", "button");
 });
 
-test.todo("works in a form as type submit");
+test("works in a form as type submit and passes the submitter", async () => {
+	const page = render(html`
+		<form>
+			<w-button name="intent" type="submit" value="save" variant="secondary">
+				Save
+			</w-button>
+		</form>
+	`);
+
+	const form = document.querySelector("form") as HTMLFormElement;
+	const wButton = document.querySelector("w-button") as HTMLElement;
+	const onSubmit = vi.fn();
+	let submitter: HTMLElement | null = null;
+	let submitContext: SubmitEvent["warpSubmitContext"];
+	let submittedIntents: FormDataEntryValue[] = [];
+
+	form.addEventListener("submit", (event) => {
+		event.preventDefault();
+
+		submitter = (event as SubmitEvent).submitter as HTMLElement | null;
+		submitContext = (event as SubmitEvent).warpSubmitContext;
+		submittedIntents = new FormData(form).getAll("intent");
+		onSubmit();
+	});
+
+	await page.getByRole("button", { name: "Save" }).click();
+
+	expect(onSubmit).toHaveBeenCalledTimes(1);
+	expect(submitter).toBeNull();
+	expect(submitContext?.initiator).toBe(wButton);
+	expect(submitContext?.nativeSubmitter).toBeNull();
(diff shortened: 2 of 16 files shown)


wkillerud:
What is the benefit of adding `warpSubmitContext` in addition to the native submitter?

digitalsadhu:
> What is the benefit of adding `warpSubmitContext` in addition to the native submitter?

After discussion with @storsletten and then subsequent testing, we found that unfortunately custom elements cannot be used as submitters. Only native elements can. Theres discussion around an API for this out there but for now, nothing on the horizon. Best we could come up with was to add a property to the event with a context object that includes the custom element that was the submitter.

wkillerud:
Sorry for being slow and difficult 🙈 I want to make sure I understand the use-case and what other options we've considered before we commit to this custom behavior "for ever".

> When Warp form components (textfield, select, checkbox, etc.) trigger implicit form submission via Enter key, the native SubmitEvent.submitter is not populated because form.requestSubmit() is called without
a submitter argument. This breaks scenarios where server frameworks or client code rely on the submitter's name/value to determine intent (e.g., which button was "clicked").


> unfortunately custom elements cannot be used as submitters. Only native elements can.

MDN and the spec has slightly different wording on the `SubmitEvent.submitter` property. I believe from reading [the spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission) that what is expected is _not_ to set the `submitter` property to the input field / `this`, but to find the form's submit button and use that as the submitter.

> If the user agent supports letting the user submit a form implicitly (for example, on some platforms hitting the "enter" key while a text control is focused implicitly submits the form), then doing so for a form, whose default button has activation behavior and is not disabled, must cause the user agent to fire a click event at that default button.

I checked out [Web Awesome's approach](https://github.com/shoelace-style/webawesome/blob/next/packages/webawesome/src/internal/submit-on-enter.ts) which is to look for a native HTML button or input of type submit in the form and use that as the submitter.

Now if the submit button is a `w-button` then we're back where we started (Web Awesome falls back to clicking on it). I think it might be a reasonable workaround to have an invisible `<input type="submit">` before `<w-button>` if you rely on having a valid `submitter`. What do you think?

storsletten:
How about making the button component create an invisible native button type submit if the consumer sets type submit on the button component. Relevant attributes such as name, value, and form-related attributes must then be mirrored from the component to the native element. I believe implicit form submission will then automatically use that native submit button even if it's hidden, and if end-user activates the button component then it can simply apply that associated native button as submitter with the submit request.

Such a workaround would also need to be clearly documented, since consumers cannot expect the object reference of the button component to match with event.submitter, but at least they can match by the name and value attributes. The name and value attributes are standardized, whereas warpSubmitContext is not, so I can see the wisdom of that approach. I believe the native button element will need to live outside shadow root though in order to participate in implicit form submissions. There may also be edge cases I haven't thought about so we really should do some thorough testing to make sure there are no unexpected snags with this approach.

A slightly modified idea is that the native button element is created only temporarily on demand when the button component needs a submitter for a submit request, but then we would need to handle implicit form submission manually, and I guess that just makes things even more complicated.

PR #752, opened 2026-08-12
feat(box): add styling api

digitalsadhu:
## Intent

Add a style API for `w-box` consistent with the approach used by other components with documented styling surfaces.

This PR removes the UnoCSS/class-based styling implementation from `w-box` and replaces it with a documented style API built on:

- component tokens (`--w-c-box-*`)
- shadow parts (`::part(base)`)

It makes `w-box` easier for consumers to style without depending on internal utility classes, while preserving the existing public boolean API for visual states like `info`, `neutral`, `bordered`, and `bleed`.

## Approach

The implementation refactors `w-box` to style from `packages/box/styles.ts` using Lit `css` instead of generated utility CSS.

Key design decisions:

- Removed `@warp-css` and `classNames` usage from `w-box`
- Added `part="base"` to the internal wrapper so consumers can override styling with `::part(base)`
- Introduced component tokens for the main surface and layout properties:
  - `--w-c-box-bg`
  - `--w-c-box-border-color`
  - `--w-c-box-border-radius`
  - `--w-c-box-border-width`
  - `--w-c-box-padding`
  - `--w-c-box-bleed-margin-inline`
- Moved visual state styling to host selectors:
  - `:host([info])`
  - `:host([neutral])`
  - `:host([bordered])`
  - `:host([bleed])`
- Kept the existing boolean state API intact instead of redesigning it into a single variant model
- Preserved the slotted last-child margin reset behavior
- Added styling API documentation in `packages/box/docs/styling.md`
- Added a Storybook `StylingApi` example to demonstrate tokens and `::part(base)`
- Added tests that verify the styling API via actual computed styles

## Risk

Main risks are visual regressions in `w-box` styling:

- default spacing, border radius, background, or border behavior could differ from the previous utility-class output
- bleed behavior could regress at narrow widths or at the `480px` breakpoint where margins/radius are restored
- combinations of boolean states like `bordered + neutral` or `bordered + info` could resolve differently if selector ordering changes in the future
- consumers depending on internal implementation details like generated classes would break, although those were not part of the intended public API

Affected surface area is limited to `w-box` consumers and layouts that depend on its current visual rendering.

## Verification

- [x] Tests added or updated
- [x] Existing tests pass
- [x] Manual verification completed, if relevant
- [ ] Screenshots or recordings attached, if UI changed

Commands run:

    pnpm vitest packages/box/box.test.ts

## AI involvement

Yes. AI was used to help implement the `w-box` style API refactor, including:

- converting the component from UnoCSS/class-based styling to CSS variable and part-based styling
- moving visual state styling to host selectors
- adding styling API docs
- adding a Storybook styling example
- adding computed-style tests for component tokens and `::part(base)`

Personally verified:

- the final code changes in `box.ts`, `styles.ts`, docs, stories, and tests
- that the component no longer depends on generated utility classes
- that the `base` part and documented component tokens are present
- that the updated `w-box` tests pass via `pnpm vitest packages/box/box.test.ts`

### DIFF
--- a/packages/box/box.ts
+++ b/packages/box/box.ts
@@ -1,7 +1,4 @@
-// @warp-css;
-
-import { classNames } from "@chbphone55/classnames";
-import { css, html, LitElement, PropertyValues } from "lit";
+import { html, LitElement, PropertyValues } from "lit";
 import { property } from "lit/decorators.js";
 import { FormControlMixin } from "@open-wc/form-control";
 
@@ -54,18 +51,7 @@ class WarpBox extends FormControlMixin(LitElement) {
 	// ::slotted([Simple Selector]) confirms to Specificity rules, but (being simple) does not add weight to lightDOM skin selectors,
 	// so never gets higher Specificity. Thus in order to overwrite style linked within shadowDOM, we need to use !important.
 	// https://stackoverflow.com/a/61631668
-	static styles = [
-		reset,
-		styles,
-		css`
-			:host {
-				display: block;
-			}
-			::slotted(:last-child) {
-				margin-bottom: 0px !important;
-			}
-		`,
-	];
+	static styles = [reset, styles];
 
 	connectedCallback(): void {
 		super.connectedCallback();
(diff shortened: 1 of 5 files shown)


packages/box/box.ts
```diff
@@ -54,18 +51,7 @@ class WarpBox extends FormControlMixin(LitElement) {
 	// ::slotted([Simple Selector]) confirms to Specificity rules, but (being simple) does not add weight to lightDOM skin selectors,
 	// so never gets higher Specificity. Thus in order to overwrite style linked within shadowDOM, we need to use !important.
 	// https://stackoverflow.com/a/61631668
```
wkillerud: Could you move this comment to the stylesheet so we remember why we need `!important`?

wkillerud (commented):
LGTM once the comment is moved

PR #753, opened 2026-08-12
fix: use correct background color token on buttons

digitalsadhu:
Hotfix for button background colors which were incorrect after a recent refactor introduced a regression.


imprashast (approved):
LGTM 👍

PR #755, opened 2026-08-13
feat(expandable): expose wrapper div as the "wrapper" CSS part

BalbinaK:
# Summary

Allows consumers to style the root wrapper and toggle button from outside the shadow DOM, e.g. `<Expandable className="part-[wrapper]:s-bg part-[button]:px-0" />`.

## Intent

Currently the Expandable of `box` type is the only one that takes full width of its parent and places the chevron at the end. However, that type has a subtle background colour, which doesn't fit e.g. the [following design:](https://www.figma.com/design/eghm57Mjb0JEzqzhucwt6G/Mobility---Item---Dealer-Page?node-id=5870-47270&m=dev)
Screenshot of an accordion section with white background

With this change the Expandable component can be used to implement the above design, preventing users from building their own components from scratch just to fix the background colour and title padding.

## Verification

- [x] Tests added or updated
- [x] Existing tests pass

### DIFF
--- a/packages/expandable/expandable.test.ts
+++ b/packages/expandable/expandable.test.ts
@@ -19,6 +19,21 @@ test("renders the given title prop and hides the slotted content", async () => {
 		.not.toBeVisible();
 });
 
+test("exposes the wrapper and toggle button as CSS parts", async () => {
+	const component = html`
+		<w-expandable title="I'm expandable">
+			<p>with expanded content</p>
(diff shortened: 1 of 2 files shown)


digitalsadhu:
@wkillerud I'm working on additions to this so you can hold of review for now.

wkillerud (approved):

